PR #252, opened 2019-11-21
improve disposer implementation

FMS-Cat:
- enable `vrm.dispose()` on dnd examples
- Refactor `disposer` a bit

### DIFF
--- a/src/vrm/utils/disposer.ts
+++ b/src/vrm/utils/disposer.ts
@@ -1,26 +1,30 @@
 import * as THREE from 'three';
 
-function disposeMaterial(material: any): void {
+function disposeMaterial(material: THREE.Material): void {
   Object.keys(material).forEach((propertyName) => {
-    if (!!material[propertyName] && typeof material[propertyName].dispose === 'function') {
-      material[propertyName].dispose();
+    const value = (material as any)[propertyName];
+    if (value && typeof value === 'object' && value.isTexture) {
+      const texture = (material as any)[propertyName] as THREE.Texture;
+      texture.dispose();
     }
   });
 
   material.dispose();
-  material = undefined;
 }
 
-function dispose(object3D: any): void {
-  if (object3D.geometry) {
-    object3D.geometry.dispose();
-    object3D.geometry = undefined;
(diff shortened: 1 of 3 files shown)


src/vrm/utils/disposer.ts
```diff
@@ -1,26 +1,32 @@
...
+
 import * as THREE from 'three';
 
-function disposeMaterial(material: any): void {
+function disposeMaterial(material: THREE.Material): void {
   Object.keys(material).forEach((propertyName) => {
-    if (!!material[propertyName] && typeof material[propertyName].dispose === 'function') {
-      material[propertyName].dispose();
+    const value = (material as any)[propertyName];
+    if (value && typeof value === 'object' && value.isTexture) {
```
yomotsu: Can we ommit `typeof value === 'object'` ?
```suggestion
    if (value && value.isTexture) {
```

src/vrm/utils/disposer.ts
```diff
@@ -1,26 +1,32 @@
...
+      const texture = (material as any)[propertyName] as THREE.Texture;
```
yomotsu: maybe?
```suggestion
      const texture = value as THREE.Texture;
```

src/vrm/utils/disposer.ts
FMS-Cat: might be? I have no clue that it is omittable

src/vrm/utils/disposer.ts
FMS-Cat: definitely

src/vrm/utils/disposer.ts
yomotsu: I think it's okay.
The line of the newer code is doing just like an optional-chaining. I think.

FMS-Cat: I just had no confidence I can index every truthy.
I know I cannot index `undefined` and `null`

FMS-Cat: yes I think I can everything other than `undefined` or `null` , my bad
I need the optional chaining so bad...... 😩

yomotsu:
LGTM!

PR #253, opened 2019-11-21
MToon: improve normal implementaiton

### LINKED ISSUE #250: MToonMaterial.toJSON does not work properly
`MToonMaterial.toJSON` has to be worked properly to use [Three.js DevTools](https://chrome.google.com/webstore/detail/threejs-developer-tools/ebpnegggocnnhleeicgljbedjkganaek)

image

FMS-Cat:
- Fixed `MToonMaterial.toJSON` (will close #250)
- MToonMaterial can now be worked with ObjectSpaceNormals
  - `bumpScale` is now just an accessor, it now uses `normalScale` instead
  - `normalMapType` will be transfered to shader as preprocessors via internal code of Three.js. Don't worry about it
- Caught up with the latest normal map implementaiton
  - See: https://github.com/mrdoob/three.js/pull/17586, https://github.com/mrdoob/three.js/pull/17598, and https://github.com/mrdoob/three.js/pull/17831


FMS-Cat:
@yomotsu could I ask you review the TS part? I have 90% confidence the shader code is good though

yomotsu:
Sorry for the delay. I have been busy lately because of end of year😢
And absolutely! I'm still watching the repo! I'll take a look

src/vrm/material/MToonMaterial.ts
```diff
@@ -110,8 +112,9 @@ export class MToonMaterial extends THREE.ShaderMaterial {
   public mainTex_ST: THREE.Vector4 = new THREE.Vector4(0.0, 0.0, 1.0, 1.0); // _MainTex_ST
   public shadeTexture: THREE.Texture | null = null; // _ShadeTexture
   // public shadeTexture_ST: THREE.Vector4 = new THREE.Vector4(0.0, 0.0, 1.0, 1.0); // _ShadeTexture_ST (unused)
-  public bumpScale = 1.0; // _BumpScale
   public normalMap: THREE.Texture | null = null; // _BumpMap. again, THIS IS _BumpMap
+  public normalMapType: THREE.NormalMapTypes = THREE.TangentSpaceNormalMap; // Three.js requires this
+  public normalScale: THREE.Vector2 = new THREE.Vector2(1.0, 1.0); // _BumpScale, in Vector2
```
yomotsu: Is the type omittable? (if so, we can do the same for `normalMapType` ) and others
```suggestion
  public normalScale = new THREE.Vector2(1.0, 1.0); // _BumpScale, in Vector2
```

FMS-Cat: sounds good

FMS-Cat: It's weird that we cannot lint this using `@typescript-eslint/no-inferrable-types`

FMS-Cat: Addressed in https://github.com/pixiv/three-vrm/pull/253/commits/65b2ab1717c0a7d49312a175f65c25b694131b9b

PR #267, opened 2019-11-26
fix GLTF

tfuru:


FMS-Cat:
I think I got what you are trying to do. It could not being built on my machine though...

FMS-Cat:
@tfuru ping

FMS-Cat:
moving to #291

PR #291, opened 2019-12-17
Better gltf import

FMS-Cat:
branched from #267 

- it now imports `GLTF` from `three/examples/jsm/loaders/GLTFLoader`


FMS-Cat:
I just have confirmed that this is actually a problem. merging

PR #377, opened 2020-03-02
GLTF.scene is now okay to be a THREE.Group instead of THREE.Scene

FMS-Cat:
`GLTF.scene` is going to be `THREE.Group` in r114
This PR make it tolerable to be used in r114.

See: https://github.com/mrdoob/three.js/pull/18601


yomotsu:
I'll take a look 👀

PR #387, opened 2020-03-17
[WIP] fix outlineWidthWorld

FMS-Cat:
outline width under `outlineWidthMode === WorldCoordinates` seems to be different a bit in three-vrm, being affected by its object scale.
This PR fixes this. Now the code is almost identical to the one in the original MToon code.

See: https://github.com/Santarh/MToon/blob/43b8a33053d59aba7efa44efd3214ba741d0d871/MToon/Resources/Shaders/MToonCore.cginc#L87-L89


FMS-Cat:
It seems there are issues on certain examples... investigating

PR #399, opened 2020-04-03
MToon: fix rimTexture and uvAnimMaskTexture

FMS-Cat:
rimTexture and uvAnimMaskTexture has malfunctioned because of ST values in vectorProperties.
This PR will fix this.


FMS-Cat:
The result of CI on GitHub is being weird but it's passed. merging this

PR #459, opened 2020-07-28
MToon: fix the direction of uv animation

### LINKED ISSUE #458: MToon: The direction of y-axis uv scroll is opposite

FMS-Cat:
Will resolve #458

The diff explains everything


FMS-Cat:
We should explicitly define the direction of uv scrolls in the spec of MToon.

Note that the spec of glTF says that the origin of the UV coordinates corresponds to the upper left of a texture image.

Ref: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#images
Ref: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureCoordinateTest

Anyway, we are going to follow the implementation in the original [MToon](https://github.com/Santarh/MToon) for now
(which scrolls backward(?) in s axis, forward(?) in t axis).

PR #478, opened 2020-09-11
VRMSpringBone: introducing center

FMS-Cat:
- "Center" feature is now functional as UniVRM
    - `center` is set for each spring bones, defines a space for the calculation of spring bones
- 🚨 It now deserializes `center` of spring bones
    - If you need the previous behavior, call `vrm.springBoneManager.setCenter(null)` after you import VRMs


FMS-Cat:
The implementation of `Matrix4WithInverseCache` is... kind of chaotic. I'm not sure the way how I did this is okay

FMS-Cat:
> plementation of `Matrix4WithInverseCache` is... kind of chaotic. I'm not sure the way how I did this is okay

have discussed about it on Discord, decided that the way is valid enough. merging

PR #482, opened 2020-09-15
Introducing VRMMeta, has an ability to load its thumbnail texture

FMS-Cat:
- `vrm.meta` is now an interface `VRMMeta` instead of `VRMSchema.Meta`
- 🚨 Interface of `vrm.meta.texture` has been changed (`number` -> `THREE.Texture`)
- 🚨 Importing a VRM from a glTF now loads a thumbnail texture by default!
    - Which may cause you unnecessary memory consumption
    - Set `VRMMetaImporter.ignoreTexture = true` to disable this behavior
- ✨ now you can extract thumbnail as a Blob using `VRMUtils.extractThumbnailBlob(renderer, vrm, resolution)`
    - See the `meta.html` example

### DIFF
--- a/examples/meta.html
+++ b/examples/meta.html
@@ -19,11 +19,17 @@
 				position: absolute;
 				background: #fff;
 			}
+			#thumbnail {
+				position: absolute;
+				top: 0;
+				right: 0;
+			}
 		</style>
 	</head>
 
 	<body>
 		<span id="meta"></span>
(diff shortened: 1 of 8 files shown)


FMS-Cat:
✨ now you can extract thumbnail as a Blob using `VRMUtils.extractThumbnailBlob(renderer, vrm, resolution)`

image

PR #506, opened 2020-10-19
WIP: Moving from webpack to rollup

### LINKED ISSUE #502: ESM
Since iife examples of Three.js are going to be deprecated, we have to find a way to integrate our modules with their esm modules (yes, we have a required dependency in the examples, GLTFLoader)

0b5vr:
will close #502 

- Changed the bundler, from webpack to rollup
- 🚨 Built file is now esm instead of umd
    - You are going to need [`es-module-shims`](https://github.com/guybedford/es-module-shims) if you want to use this without bundlers! See the updated README.md for details
- Updated README and examples accordingly

### Need to check

- [ ] Does this work using npm using bundlers?
- [ ] Need to check the spec of import-maps again


FMS-Cat:
There's a lot of conflict! I think it'd better doing the same thing from scratch. closing

PR #507, opened 2020-10-20
fix (humanoid): getPose / setPose is now working as intended

### LINKED ISSUE #503: model.humanoid.setPose(model.humanoid.restPose) Bug?
I don't know if is this a bug or I misused this function, when I call `model.humanoid.setPose(model.humanoid.restPose)` to reset the VRM model's pose, the model is getting weird.
image

After I dig in the code, I think the reason causing this bug is in here, which the original bone position is added to the pose data passed by the argument, then the position of all bones are doubled: https://github.com/pixiv/three-vrm/blob/1b4aa0a72b7153a6fba5bc575932f7a8dcb807b8/src/vrm/humanoid/VRMHumanoid.ts#L86-L93

If this is not a bug, is there any proper way to reset the model to its initial state?

FMS-Cat:
Should resolve #503 

- 🚨💡 Greatly changed the behavior of `getPose` / `setPose`
    - It now explicitly receives / returns local, relative transforms that is described in `VRMPose`
- ✨ New method: `resetPose`
    - It just resets the pose to its restPose (that should be T stance)
- 🚨💡 `restPose` is now explicitly not compatible with `getPose` / `setPose`
    - If you want to reset its pose, I have introduced `resetPose` so use this instead

### DIFF
--- a/src/vrm/humanoid/VRMHumanoid.ts
+++ b/src/vrm/humanoid/VRMHumanoid.ts
@@ -130,6 +130,13 @@ export class VRMHumanoid {
   }
 
   /**
+   * Reset the humanoid to its rest pose.
+   */
+  public resetPose(): void {
+    this.setPose({});
+  }
+
+  /**
    * Return a bone bound to a specified [[HumanBone]], as a [[VRMHumanBone]].
    *
    * See also: [[VRMHumanoid.getBones]]
--- a/src/vrm/humanoid/VRMHumanoid.ts
+++ b/src/vrm/humanoid/VRMHumanoid.ts
@@ -27,7 +27,7 @@ export class VRMHumanoid {
    * A [[VRMPose]] that is its default state.
    * Note that it's not compatible with `setPose` and `getPose`, since it contains non-relative values of each local transforms.
    */
-  private readonly _restPose: VRMPose = {};
+  public readonly restPose: VRMPose = {};
 
   /**
    * Create a new [[VRMHumanoid]].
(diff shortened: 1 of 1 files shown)


src/vrm/humanoid/VRMHumanoid.ts
```diff
@@ -55,9 +59,33 @@ export class VRMHumanoid {
...
+        _v3A.fromArray(restState.position).negate();
+      }
+      if (restState?.rotation) {
+        _quatA.fromArray(restState.rotation).inverse();
+      }
+
+      // 🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥
+      if (!restState) {
+        if (vrmBoneName === VRMSchema.HumanoidBoneName.LeftUpperArm) {
+          node.rotation.x = 0.5 * Math.PI;
```
yomotsu, line 78: Should this be something like `PI_HALF` constant? like
https://github.com/pixiv/three-vrm/blob/f10623dbcf7470624dafbc6bb10302f8267d2308/src/vrm/material/MToonMaterial.ts#L8

FMS-Cat: oh no

it's just a leftover code. I completely forgot to remove despite I've put a lot of 🔥 on it. Let me just remove these lines 🙇 🙇 🙇

FMS-Cat:
@JLChnToZ Does this look good for you?

JLChnToZ:
Yeah I think so, thanks a lot!

yomotsu (approved):
Looks Good❤️

JLChnToZ (approved):

FMS-Cat:
thanks! gonna merge

PR #511, opened 2020-10-26
Add OffscreenCanvas support for extractThumbnailBlob

### LINKED ISSUE #508: VRMUtils.extractThumbnailBlob does not work with OffscreenCanvas backed WebGLRenderer
Hi, I have find that if I initialize the renderer with an OffscreenCanvas such as moving the rendering logic to a webworker, I cannot use `VRMUtils.extractThumbnailBlob` directly.

Currently I'm using a workaround like this, which added some extra flow to the original code:
```typescript
export function extractThumbnailBlob(renderer: WebGLRenderer, vrm: VRM, size = 512): Promise<Blob> {
  // get the texture
  const texture = vrm.meta?.texture;
  if (!texture) {
    throw new Error('extractThumbnailBlob: This VRM does not have a thumbnail');
  }

  // store the current resolution
  renderer.getSize(_v2A);
  const prevWidth = _v2A.x;
  const prevHeight = _v2A.y;

  // overwrite the resolution
  // OffscreenCanvas does not have any styling method, it will stuck here if not handling with care.
  renderer.setSize(size, size, !(renderer.getContext().canvas instanceof OffscreenCanvas));

  // assign the texture to plane
  _plane.material.map = texture;

  // render
  renderer.render(_scene, _camera);

  // get blob
  return new Promise((resolve, reject) => {
    const { canvas } = renderer.getContext();
    // The method of converting rendered data to blob is a bit different between HTMLCanvasElement and OffscreenCanvas.
    if (canvas instanceof OffscreenCanvas) {
      canvas.convertToBlob().finally(() => {
        // revert to previous resolution
        renderer.setSize(prevWidth, prevHeight, false);
      }).then(resolve, reject);
    } else {
      canvas.toBlob((blob) => {
        // revert to previous resolution
        renderer.setSize(prevWidth, prevHeight);

        if (blob == null) {
          reject('extractThumbnailBlob: Failed to create a blob');
        } else {
          resolve(blob);
        }
      });
    }
  });
}
```

JLChnToZ:
fixes #508

### DIFF
--- a/src/vrm/VRMUtils/extractThumbnailBlob.ts
+++ b/src/vrm/VRMUtils/extractThumbnailBlob.ts
@@ -25,13 +25,15 @@ export function extractThumbnailBlob(renderer: THREE.WebGLRenderer, vrm: VRM, si
     throw new Error('extractThumbnailBlob: This VRM does not have a thumbnail');
   }
 
+  const canvas = renderer.getContext().canvas;
+
   // store the current resolution
   renderer.getSize(_v2A);
   const prevWidth = _v2A.x;
   const prevHeight = _v2A.y;
 
   // overwrite the resolution
-  renderer.setSize(size, size);
+  renderer.setSize(size, size, !(canvas instanceof OffscreenCanvas));
 
   // assign the texture to plane
   _plane.material.map = texture;
@@ -43,8 +45,15 @@ export function extractThumbnailBlob(renderer: THREE.WebGLRenderer, vrm: VRM, si
   _plane.material.map = null;
 
   // get blob
+  if (canvas instanceof OffscreenCanvas) {
+    return canvas.convertToBlob().finally(() => {
+      // revert to previous resolution
+      renderer.setSize(prevWidth, prevHeight, false);
+    });
+  }
+
   return new Promise((resolve, reject) => {
-    (renderer.getContext().canvas as HTMLCanvasElement).toBlob((blob) => {
+    canvas.toBlob((blob) => {
       // revert to previous resolution
       renderer.setSize(prevWidth, prevHeight);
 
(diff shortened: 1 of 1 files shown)


src/vrm/VRMUtils/extractThumbnailBlob.ts
```diff
@@ -25,13 +25,15 @@ export function extractThumbnailBlob(renderer: THREE.WebGLRenderer, vrm: VRM, si
...
+  const canvas = renderer.getContext().canvas;
+
   // store the current resolution
   renderer.getSize(_v2A);
   const prevWidth = _v2A.x;
   const prevHeight = _v2A.y;
 
   // overwrite the resolution
-  renderer.setSize(size, size);
+  renderer.setSize(size, size, !(canvas instanceof OffscreenCanvas));
```
FMS-Cat, line 36: Oh! I didn't know the third argument of setSize is `true` by default.

There is no reason resizing the style in this function, could you set false here instead? I also want to change `revert to previous resolution` one of course.

FMS-Cat:
I've never used offscreen canvas before but let me try this in this opportunity...

src/vrm/VRMUtils/extractThumbnailBlob.ts
JLChnToZ: Do you mean all `renderer.setSize()` in this method do not need to updateStyle?

FMS-Cat:
It worked perfectly!

src/vrm/VRMUtils/extractThumbnailBlob.ts
FMS-Cat: yes

FMS-Cat:
uhhhh I just have noticed that the resize applies to the screen buffer for several frames since `canvas.toBlob` is asynchronous @ `examples/meta.html` ......
I don't think we should address to this in this branch but I'll think something against the issue.

FMS-Cat:
Anyway it looks excellent for me, thanks again! merging

FMS-Cat (approved):

PR #521, opened 2020-11-04
Make it monorepo

FMS-Cat:
We are going to separate some components into multiple modules when we do VRM 1.0 support.
This PR makes the project able to manage multiple modules using monorepo, there's only one package available at this point.


FMS-Cat:
it successfully publishes! merging

PR #527, opened 2020-11-04
WIP: Moving from webpack to rollup, attempt 2

### LINKED ISSUE #144: `README.md` npm example's module usage is confusing
The module usage of the "npm" example in `README.md` is confusing. The example has the following `import` statements:
```JavaScript
import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader';
import { VRM } from '@pixiv/three-vrm';
```

However, `GLTFLoader` and `VRM` is apparently referred as `THREE.GLTFLoader` and `THREE.VRM`. Then what does the literal `GLTFLoader` and `VRM` does?

Also, it does not tell how `import` should be interpreted. It matters because there are number of options for interpretation (e.g. the "babel" transpiler, webpack, and Node.js's ESM.)

### LINKED ISSUE #502: ESM

0b5vr:
will close #502 , will close #144
Retry of #506 

- Changed the bundler, from webpack to rollup
- 🚨 `.module.js` is now esm instead of umd
    - ~~You are going to need [`es-module-shims`](https://github.com/guybedford/es-module-shims) if you want to use module build without bundlers! See the updated README.md for details~~
        - We are going to continue providing UMD versions for now. See: mrdoob/three.js#20568
- Updated README and examples accordingly
- In IIFE variant, VRM components are installed under `THREE_VRM`
    - Also be installed under `THREE` as previous

### Need to check

- [ ] Does this work with npm using bundlers?
- [x] ~~Need to check the spec of import-maps again~~
    - We don't have to secure this for now because its still drafter than we thought. See: mrdoob/three.js#20568
- [x] How about separated components in future?
    - Monorepo seems to be working good


FMS-Cat:
It seems we cannot drop iife and cjs output yet...... 😩 

https://github.com/mrdoob/three.js/pull/20568

PR #534, opened 2020-11-06
Move schema to individual package

0b5vr:
### ~~Why DO NOT MERGE~~

- ~~The breaking change is too big, we should wait till VRM1.0 support~~

### About

- This PR provides schema of VRM as individual packages
- ~~🚨: We are going to no longer be not able to use `VRMSchema.HumanoidBoneName` and `VRMSchema.BlendShapePresetName` as enums~~
    - ~~Since we're going to make them union types instead of enums~~
    - ~~We should provide a deprecation warning and compat for a while though, it's too terrible to disable it immediately~~
    - **Update:** union type + const value should be included in individual implementation modules
    - **Update:** Now it should not have any changes other than adding schema packages


FMS-Cat:
Changed my mind: `VRMSchema.HumanoidBoneName` should be exist as enums since Three.js uses such kind of enums

0b5vr:
rewinded all the changes about `three-vrm`, now it should not have any changes other than adding schema packages

PR #577, opened 2020-12-26
Fix human bone assignment failure for gltf nodes[0]

saturday06:
`VRMHumanoidImporter.import()`  assigns VRM Humanoid bone name to glTF node. But glTF node with its index of zero is unexpectedly excluded from that assignment.

### DIFF
--- a/packages/three-vrm/src/vrm/humanoid/VRMHumanoidImporter.ts
+++ b/packages/three-vrm/src/vrm/humanoid/VRMHumanoidImporter.ts
@@ -30,7 +30,7 @@ export class VRMHumanoidImporter {
     if (schemaHumanoid.humanBones) {
       await Promise.all(
         schemaHumanoid.humanBones.map(async (bone) => {
-          if (!bone.bone || !bone.node) {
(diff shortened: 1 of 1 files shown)


packages/three-vrm/src/vrm/humanoid/VRMHumanoidImporter.ts
```diff
@@ -30,7 +30,7 @@ export class VRMHumanoidImporter {
     if (schemaHumanoid.humanBones) {
       await Promise.all(
         schemaHumanoid.humanBones.map(async (bone) => {
-          if (!bone.bone || !bone.node) {
+          if (!bone.bone || typeof bone.node !== 'number') {
```
saturday06, line 33: Fix unexpected return when `bone.node` is `0`.

FMS-Cat: ohhhhhh

personally prefer this:

```suggestion
          if (!bone.bone || bone.node == null) {
```

FMS-Cat:
apparent. thanks!

PR #589, opened 2021-01-08
MToon: add a missing line to MToon Shader

0b5vr:
Because of the missing line the behavior of MToon is being slightly wrong. This PR fixes this.

The equivalent line in the original MToon: https://github.com/Santarh/MToon/blob/master/MToon/Resources/Shaders/MToonCore.cginc#L218


saturday06:
LGTM!

PR #594, opened 2021-01-13
Materials: KeywordMap is now okay to be null

0b5vr:
Seems there was a bug in UniVRM, puts `undefined` to `keywordMap` of `materialProperties` instead of an empty object (it's valid in VRM spec, by the way.)
three-vrm could not handle `materialProperties` without `keywordMap` properly, this PR fixes this.

Considering [this issue comment](https://github.com/vrm-c/UniVRM/issues/172#issuecomment-640425765), the code that uses `keywordMap` seems to be outdated, so I removed the lines.
We have to see if the behavior is even with UniVRM before we merge this.

Issue @ UniVRM: https://github.com/vrm-c/UniVRM/issues/654


saturday06:
LGTM!

PR #609, opened 2021-01-26
fix (MToon): texture transform was opposite from UniVRM's

0b5vr:
### About

Since [uv of models are opposite between UniVRM and Three.js](https://github.com/vrm-c/UniVRM/blob/v0.65.0/Assets/UniGLTF/Runtime/UniGLTF/IO/MeshImporter.cs#L166), we were miscalculating the uv transform defined by `materialProperties/i/vectorProperties/_MainTex`.
This PR will make the behavior same as UniVRM's.

This non-specification won't be took over VRM 1.0 by the way. Don't worry about it.

### Test model

Use this model to confirm the behavior: [uv-texture-transform.zip](https://github.com/pixiv/three-vrm/files/5871205/uv-texture-transform.zip)

This model contains a plane that has a UV grid texture.
The uv is transformed using `/extensions/VRM/materialProperties/0/vectorProperties/_MainTex` . The value is `[0.125, 0.125, 0.5, 0.5]` .

The model looks like this in UniVRM (The model should be rendered in this way):
image

This is the current behavior in three-vrm v0.5.6:
image

This PR will make it look like this:
image


0b5vr:
Current status: Looking for a reviewer

nyamadan:
LGTM

0b5vr:
merging

PR #650, opened 2021-03-04
fix (VRMHumanoid): make API of getBoneNodes sane

0b5vr:
### Description

- `VRMHumanoid.getBones` now returns `[]` instead of `undefined` when there are no bones bound to the specified HumanBone
- `VRMHumanoid.getBoneNodes` now returns `[]` instead of emitting an internal error (😩) when there are no bones bound to the specified HumanBone
- Trivial, change a `||` into `??`


nyamadan:
LGTM

nyamadan (approved):

PR #825, opened 2021-10-01
Bump three.js to r133

0b5vr:
- Bump three.js to r133
    - https://github.com/mrdoob/three.js/releases/tag/r133
- I had to fix a shader in order to make it work
    - https://github.com/mrdoob/three.js/pull/22471


0b5vr:
All examples are working without any problem. merging

0b5vr:
wait, I might want to implement a compat for the shader change instead

PR #840, opened 2021-10-20
fix (VRM1.0, mtoon): fix mtoon when there are multiple lights

0b5vr:
Certain variables related to lighting are defined inside of unroll loop which causes redifinition errors,
and it only happens when there are multiple lights.
This PR will fix this.

Redo of #801 in 1.0 branch


0b5vr:
forgot to merge this,,,

PR #811, opened 2021-09-10
Bump three.js to r132

0b5vr:
will resolve #809 

This PR will bump three.js to r132.
It also includes fixes of shader compat break @ r132

PR #841, opened 2021-10-21
[1.0] bump three.js to r133

0b5vr:
### Description

This PR will bump three.js from r126 to r133.
There was a lot of issues caused by this bump, mostly on MToonMaterial and its shader.
That was a longer journey than I thought,,,

The part of it is redo of #811 and #825 on `1.0` branch.
See commit log for details if you are interested.


0b5vr:
I think it's working on r126-r133.
There is a low priority issue though (#842).
Merging

PR #851, opened 2021-11-02
[1.0] Fix rim lighting

0b5vr:
- Beginning from Three.js r133, it utilizes `SRGB8_ALPHA8` textures instead of inline conversions. This PR supports the behavior.
    - See: https://github.com/mrdoob/three.js/pull/22551
    - Because of this, procedures in `_updateShaderCode` are replaced by `onBeforeCompile` .
- MatCap was not working as "MatCap", this PR should fix the behavior.


0b5vr:
Confirmed that it's working between r126-r133, merging

PR #852, opened 2021-11-04
[1.0] Fix mtoon cache key

0b5vr:
`MToonMaterial.customProgramCacheKey` is needed to keep it update using `onBeforeCompile`


0b5vr:
- Tested on bunch of models
- Every examples are working properly on three-vrm-girl-1.0.
- `materials-debug.html` is working on r126-r133

merging

PR #880, opened 2021-12-28
Remove unnecessary vertices

0b5vr:
### Description

Add a new utility function `VRMUtils.removeUnnecessaryVertices`.

This is to address the issue that morph textures consumes gigantic amount of VRAM.
See comments for more details.

I probably have to apply this also for the current v0.x stream.

See: mrdoob/three.js#23095

### Points need review

- [ ] TBD

PR #887, opened 2022-01-12
fix: removeUnnecessaryVertices should clone boundingBox and boundingSphere

0b5vr:
### Description

removeUnnecessaryVertices (#880) was not cloning `boundingBox` and `boundingSphere` of geometries. This PR will fix this.
I also made this copy `name` and `userData` .


0b5vr:
I believe the diff is obvious + I've tested on our production code. Merging by myself

PR #888, opened 2022-01-12
[1.0] type GLTFSchema using gltf-transform

0b5vr:
### Description

- Type the schema of glTF using [gltf-transform](https://github.com/donmccurdy/glTF-Transform). Shoutouts to donmccurdy!
    - I've caught several issues with this type on loader plugins, so I fixed them.
- about `gltfExtractPrimitivesFromNode` ...
    - it is no longer used in `three-vrm` , so I've deleted the one.
    - The one defined in `three-vrm-core` was the older one! I've replaced the one with the newer one.

### Points need review

- [ ] is the typing look good?
    - especially about optional types and assertions
- [ ] I have 80% confidence the diff is typo-free
    - Please review the rest 20%,,,,,,


nyamadan (approved):
LGTM

PR #895, opened 2022-01-14
Bump Three.js to r136

0b5vr:
### Description

- Bump three.js to r136.
- There was a breaking change about color encoding and the forked `getTexelDecodingFunction` affects the change. Fixed this.


0b5vr:
seems it's working fine!

PR #913, opened 2022-01-28
[1.0] Bump Three.js to r137

0b5vr:
Redo of #910 

- Bump Three.js to r137.
- Add `.js` to examples imports
- Remove shader side encoding functions
    - See: https://github.com/mrdoob/three.js/pull/23129


0b5vr:
gonna include compat changes I've done in #916

PR #942, opened 2022-03-10
fix (1.0, schema): fix signatures of `extensions`

0b5vr:
### Description

Changed type signature of `.extensions` of gltf properties from `{ [key: string]: { [key: string]: any } }` to `{ [name: string]: any }` .

See: https://github.com/pixiv/three-vrm/pull/934#discussion_r822597191

### Points need review

- I think it's obvious based on the previous discussion


nyamadan (approved):
よさそうです！

PR #954, opened 2022-03-30
fix: add compatibility for `getHemisphereLightIrradiance`

### LINKED ISSUE #938: Adding HemisphereLight causes shader compile error
(Tested in examples/dnd.html)
Adding HemisphereLight like this `scene.add(new THREE.HemisphereLight());` causes compiler error as following:
```
THREE.WebGLProgram: Shader Error 1282 - VALIDATE_STATUS false
Program Info Log: Fragment shader is not compiled.
FRAGMENT
ERROR: 0:1037: 'getHemisphereLightIrradiance' : no matching overloaded function found
```
I'm not familiar with shader, but apparently the type of the parameter passed to getHemisphereLightIrradiance is not correct.
Should it be geometry.normal, just like in getLightProbeIrradiance?
I have tried it and it doesn't give error anymore but I don't know if it's correct to pass geometry.normal.
https://github.com/pixiv/three-vrm/blob/e2ce97021d8c92686081d7208c634efdbae29ac5/packages/three-vrm/src/material/shaders/mtoon.frag#L548-L560

you74674:
Fix #938 
According to #825, I think the change(check version>=133 and pass geometry.normal) should be correct.

### DIFF
--- a/packages/three-vrm/src/material/shaders/mtoon.frag
+++ b/packages/three-vrm/src/material/shaders/mtoon.frag
@@ -554,7 +554,11 @@ void main() {
   #if ( NUM_HEMI_LIGHTS > 0 )
     #pragma unroll_loop_start
     for ( int i = 0; i < NUM_HEMI_LIGHTS; i ++ ) {
-      irradiance += getHemisphereLightIrradiance( hemisphereLights[ i ], geometry );
+      #if THREE_VRM_THREE_REVISION >= 133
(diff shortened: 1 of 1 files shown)


0b5vr:
I see, I'm sorry I could not notice the issue #938 :bow:

Are you sure that the change has been made in r133?

0b5vr:
Yep, it seems to be. I just have overseen the change.

https://github.com/mrdoob/three.js/pull/22471

0b5vr (approved):
Looks good. Thank you!

PR #977, opened 2022-05-23
examples (1.0, constraint): fix aim.html

0b5vr:
Fixed an example `aim.html` of 1.0 node_constraint plugin.
The enum format of `aimAxis` was outdated


0b5vr:
The change is obvious, tested and worked fine. merging

PR #1003, opened 2022-07-06
BREAKING change (1.0, springBone): VRMSpringBoneManager, rename several members

0b5vr:
### Description

Rename several members of `VRMSpringBoneManager` .

- `springBones` -> `joints`
- `addSpringBone` -> `addJoint`
- `removeSpringBone` -> `removeJoint`

previous names still work, but emit a deprecation warning.

### Reason

"SpringBone" rather means the entire system, while "Joint" indicates the component of the SpringBone.


ke456-png (approved):
> addSpringBone() is deprecated. Use addJoint() instead.

LGTM

PR #1014, opened 2022-07-25
deps: Remove quicktype

0b5vr:
### Description

Remove `quicktype` from devDependencies.

It used to be used by three-vrm v0 for generating glTF types.

Now we are using gltf-transform instead: https://github.com/pixiv/three-vrm/pull/888


ke456-png:
LGTM

PR #1031, opened 2022-08-30
BREAKING change: VRM now requires meta and humanoid to instantiate

0b5vr:
### Description

VRM now requires meta and humanoid to instantiate

If loader could not find valid meta and humanoid, it does not create VRM object

this also fixes the bug where it loads a empty VRM class when the loaded glTF is not a VRM

VRM.meta and VRM.humanoid are no longer optional!

### DIFF
--- a/packages/three-vrm/src/VRMLoaderPlugin.ts
+++ b/packages/three-vrm/src/VRMLoaderPlugin.ts
@@ -3,8 +3,10 @@ import { GLTF, GLTFLoaderPlugin, GLTFParser } from 'three/examples/jsm/loaders/G
 import {
   VRMExpressionLoaderPlugin,
   VRMFirstPersonLoaderPlugin,
+  VRMHumanoid,
   VRMHumanoidLoaderPlugin,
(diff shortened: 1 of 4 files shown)


packages/three-vrm/src/VRMLoaderPlugin.ts
```diff
@@ -98,17 +100,25 @@ export class VRMLoaderPlugin implements GLTFLoaderPlugin {
...
-      nodeConstraintManager: gltf.userData.vrmNodeConstraintManager,
-    });
-    gltf.userData.vrm = vrm;
+    const meta = gltf.userData.vrmMeta as VRMMeta | null;
+    const humanoid = gltf.userData.vrmHumanoid as VRMHumanoid | null;
+
+    // meta and humanoid are required to be a VRM.
+    // Don't create VRM if they are null
+    if (meta && humanoid) {
+      const vrm = new VRM({
```
ke456-png, line 109: metaかhumanoidが無いときVRMは作られない。

ke456-png (approved):
LGTM

PR #1046, opened 2022-09-16
[1.0] chore: we can now specify rollup serve port using env PORT

0b5vr:
### Description

We can now specify rollup serve port using env `PORT`.

This is useful when you edit two or more packages at once.

Also, added a note about this case in CONTRIBUTING.md.


ke456-png (approved):
NICE! 👍

PR #1151, opened 2023-01-06
deps: bump Three.js to r148

0b5vr:
This PR bumps the Three.js to r148.

One notable update for us is that shadow maps support material.map with alphaTest: https://github.com/mrdoob/three.js/pull/25000


ke456-png (approved):
NICE! 👍

PR #1164, opened 2023-01-23
Draft: Fix LookAt position

0b5vr:
Fixed the behavior of offsetFromHeadBone.

offsetFromHeadBone should be a world space offset instead of the local space offset.

However, the current specification unclearly says that it should be evaluated in the local space of the head bone, which is not convenient.
VRMC is now discussing the behavior.

This PR also improves the LookAt example, added the lookAt gizmo.


0b5vr:
According to the discussion in VRMC, the current behavior seems to be correct. Closing.

We have to fix our sample model instead.

PR #1168, opened 2023-02-02
docs: Update migration-guide-1.0

0b5vr:
Add a section about the change of `VRM.materials`.

`VRM.materials` used to have `MeshStandardMaterial`s and `MeshBasicMaterial`s in three-vrm v0.x but it's not in v1.0.

This change clarifies why the change is made and note a workaround.

Discussion source: https://github.com/pixiv/three-vrm/discussions/1167#discussioncomment-4847980


ke456-png (approved):
NICE! 👍

PR #1183, opened 2023-03-03
Update sample vrm

ke456-png:
Fix lookAt offset of sample vrm.

https://github.com/vrm-c/vrm-specification/pull/427


0b5vr (approved):
Thank you!

PR #1250, opened 2023-07-11
Default shadingShiftTextureScale to 1.0

### LINKED ISSUE #1248: Default `shadingShiftTextureScale` to 1
The MToon spec indicates that `shadingShiftTextureInfo.scale` should have a default value of "1.0"
CleanShot 2023-07-10 at 12 33 04@2x

[(source permalink)](https://github.com/vrm-c/vrm-specification/blob/0861a66eb2f2b76835322d775678047d616536b3/specification/VRMC_materials_mtoon-1.0/schema/mtoon.shadingShiftTexture.schema.json)

Here is a sample file with no "scale" define for its shadingShiftTexture:
[shadingShiftTexture-no-scale.vrm.zip](https://github.com/pixiv/three-vrm/files/12004259/shadingShiftTexture-no-scale.vrm.zip)

When I load it with three-vrm [(mtoon demo)](https://pixiv.github.io/three-vrm/packages/three-vrm-materials-mtoon/examples/loader-plugin.html), the `shadingShiftTextureScale` is set to `null` and the rendered view doesn't look as expected.

CleanShot 2023-07-10 at 12 31 24@2x
CleanShot 2023-07-10 at 12 31 59@2x

Manually setting this to 1 produces the expected result:

CleanShot 2023-07-10 at 12 32 11@2x

mattrossman:
Resolves #1248

Tested with the `loader-plugin.html` example and sample file from that issue.

### DIFF
--- a/packages/three-vrm-materials-mtoon/src/MToonMaterial.ts
+++ b/packages/three-vrm-materials-mtoon/src/MToonMaterial.ts
@@ -419,7 +419,7 @@ export class MToonMaterial extends THREE.ShaderMaterial {
         shadingShiftFactor: { value: 0.0 },
         shadingShiftTexture: { value: null },
         shadingShiftTextureUvTransform: { value: new THREE.Matrix3() },
-        shadingShiftTextureScale: { value: null },
+        shadingShiftTextureScale: { value: 1.0 },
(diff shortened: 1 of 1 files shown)


ke456-png (approved):
NICE! 👍

Issue #1251, opened 2023-07-13
Could not resolve `@types/three@"^0.152.1`


marwie:
Hi all, 

I just tried installing 2.0.0 from npm but I currently get an error that it couldnt resolve the peerjs types `@types/three@"^0.152.1`
We are using `@types/three@"0.153.0` in our package right now. 

What do you think about changing the types peerDependency to something like `>= 0.152.1` ?

Related to https://github.com/pixiv/three-vrm/pull/1232

marwie:
Looks like types for 154 were just released as well https://www.npmjs.com/package/@types/three

PR #1254, opened 2023-07-18
Bump three types to 153

marwie:
see PR https://github.com/pixiv/three-vrm/issues/1251

### DIFF
--- a/packages/three-vrm-core/package.json
+++ b/packages/three-vrm-core/package.json
@@ -50,12 +50,12 @@
     "@pixiv/types-vrmc-vrm-1.0": "2.0.0"
   },
   "devDependencies": {
-    "@types/three": "^0.152.1",
+    "@types/three": "^0.153.0",
     "lint-staged": "13.1.2",
     "three": "^0.153.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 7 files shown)


0b5vr:
Didn't noticed @types/three for r153 has been released, thank you.
I will check this later.

marwie:
No problem, types for 154 are already released too - not sure if you tested the package with latest three yet

When the types errors are fixed we could start using it in our engine which uses three 154

0b5vr:
@marwie 

I will bump to r154 along with the type definition later, I'm not sure whether the library works as is or not.

I'm going to deploy the @types/three r153 changes for now.

Could you also apply the change to `yarn.lock`?

marwie:
@0b5vr done

0b5vr:
Thank you!

0b5vr (approved):

marwie:
@0b5vr let me know if/when I should test with 154. Ideally we can update the types to 154 as well to avoid the peer dependency errors altogether :)

image

0b5vr:
Yes, we definitely need to reconsider peerDependencies to change to a realistic way.

PR #1288, opened 2023-09-11
Expose `VRMExpressionBind*` classes

### LINKED ISSUE #1287: Feature request: Expose `VRMExpressionBind*` classes
I wish to programmatically add expressions to a VRM model via `VRMExpressionManager.registerExpression()`. This method expects a `VRMExpression` whose binds have been added via `VRMExpression.addBind()`

Although the `.registerExpression()` method is exposed, I cannot create binds for `.addBind()` because the following classes are not bundled:

- `VRMExpressionMaterialColorBind`
- `VRMExpressionMorphTargetBind`
- `VRMExpressionTextureTransformBind`

Only the type definitions for these are available due to the usage of `export type` in [expressions/index.ts](https://github.com/pixiv/three-vrm/blob/358ff2a3bb66568967bf54b6fb9adbb0e3b4ba6c/packages/three-vrm-core/src/expressions/index.ts)

I request that these be changed to plain `export` so users can construct new expression binds.

I can submit a PR for this.

mattrossman:
Resolves #1287


yue4u:
Thanks! I'll make a release as soon as possible.

PR #1293, opened 2023-09-22
build: bundle dts via api-extractor and move deps to dev only

### LINKED ISSUE #1289: Suggestion: Bundle `three-vrm` types, move subpackages to `devDependencies`
I faced some difficulties when trying to make a fix in `three-vrm`.

The published contents of `@pixiv/three-vrm` contained `.js` files that bundle all subpackages together, but the `.d.ts` still reference individual subpackages from `@pixiv/three-vrm-*`. Additionally, these subpackages are [listed as "dependencies"](https://github.com/pixiv/three-vrm/blob/358ff2a3bb66568967bf54b6fb9adbb0e3b4ba6c/packages/three-vrm/package.json#L49-L54) of `three-vrm` even though their code is already bundled. These dependencies are only used to provide types for the bundle.

This causes the following issues:

* Users download unused `.js` bundles for each subpackage to `node_modules`
* Type definitions for `three-vrm` can get desynchronized from the actual `.js` code

For example, I tried to fork the repo to make a small change. After building and publishing `three-vrm` to my own namespace, I experienced strange build errors. Even though my published `.js` files contained the desired changes, the types were still pointing at the unchanged upstream `three-vrm-core` package. This means I'd have to publish each subpackage individually to my namespace!

A solution I adopted in my fork is to bundle the types for `three-vrm` using [tsup](https://github.com/egoist/tsup). This means the published `.js` and `.d.ts` files are sure to be in sync, and subpackages are no longer required as "dependencies".

Here is the change: https://github.com/mattrossman/three-vrm/commit/3c1e258dea5beaa0fa5509a1407837a5f89ce40c

This might make it easier for folks in the community to make fixes, while also reducing the package's footprint in node_modules :)
This shouldn't affect users who wish to install the subpackages individually.

|`node_modules` before|`node_modules` after|
|---|---|
|CleanShot 2023-09-11 at 14 55 34@2x | CleanShot 2023-09-11 at 14 53 58@2x|

Just a suggestion—I can make a PR for this if you're interested, otherwise feel free to close.

yue4u:
close #1289

### `@microsoft/api-extractor`

- adopters:
  - vite, angular, puppeteer
- pros:
  - actively developed
  - can be used as interface checker
- cons:
  - needs configuration
  - heavy / maybe over complicated for us

### `rollup-plugin-dts`

- adopters:
  - vue, tsup 
- pros:
  - less configuration
  - faster? (no tested)
    - vue switched from `@microsoft/api-extractor` to `rollup-plugin-dts` from after 3.3
- cons:
  - generates wrong code in certain conditions https://github.com/egoist/tsup/issues?q=dts
  - currently in [maintenance-mode](https://github.com/Swatinem/rollup-plugin-dts#maintenance-mode)

### do not bundle at all

- pros:
  - less configuration
  - fast?
  - tree shaking support?
    - if we go esm only it's actually not bad to not bundle at all 
- cons:
  - esm/cjs support
  - cdn may break/hard to try out
  - large install size
## Needs discussion:
- [ ] do we need api.md?


yarn.lock
```diff
@@ -1345,9 +1423,9 @@
     meshoptimizer "~0.18.1"
 
 "@types/webxr@*":
-  version "0.5.0"
-  resolved "https://registry.yarnpkg.com/@types/webxr/-/webxr-0.5.0.tgz#aae1cef3210d88fd4204f8c33385a0bbc4da07c9"
-  integrity sha512-IUMDPSXnYIbEO2IereEFcgcqfDREOgmbGqtrMpVPpACTU6pltYLwHgVkrnYv0XhWEcjio9sYEfIEzgn3c7nDqA==
+  version "0.5.5"
+  resolved "https://registry.yarnpkg.com/@types/webxr/-/webxr-0.5.5.tgz#9e0a27e809c8f76cc1ef525d9f96b8fd94ef9c42"
```
yue4u, line 1427: modified to contain this fix https://github.com/DefinitelyTyped/DefinitelyTyped/commit/b2505fdf433904702dc8a5d6c07d06498b1a6ee5

yue4u:
https://github.com/pixiv/three-vrm/pull/1293/commits/fdb0c879ab1950ba698e0ef95206dba04dea266d addressed missing import issues. The added exports are most for building the dts bundle.

0b5vr:
the report files end with `.api.md`, are those files meant to be committed to?

packages/three-vrm/src/VRMUtils/index.ts
```diff
@@ -3,6 +3,8 @@ import { removeUnnecessaryJoints } from './removeUnnecessaryJoints';
 import { removeUnnecessaryVertices } from './removeUnnecessaryVertices';
 import { rotateVRM0 } from './rotateVRM0';
 
+export type { deepDispose, removeUnnecessaryJoints, removeUnnecessaryVertices, rotateVRM0 };
```
0b5vr, line 6: why is the line needed?

yue4u: see https://github.com/pixiv/three-vrm/pull/1293#issuecomment-1740419724

packages/three-vrm/src/index.ts
```diff
@@ -9,3 +9,6 @@ export * from '@pixiv/three-vrm-core';
 export * from '@pixiv/three-vrm-materials-mtoon';
 export * from '@pixiv/three-vrm-node-constraint';
 export * from '@pixiv/three-vrm-springbone';
+
+export type { VRMMaterialsV0CompatPlugin } from '@pixiv/three-vrm-materials-v0compat';
+export type { VRMNodeConstraintLoaderPlugin } from '@pixiv/three-vrm-node-constraint';
```
0b5vr, line 14: Doesn't it overwrapped with the line `export * from '@pixiv/three-vrm-node-constraint';` ?

packages/three-vrm/src/index.ts
```diff
@@ -9,3 +9,6 @@ export * from '@pixiv/three-vrm-core';
 export * from '@pixiv/three-vrm-materials-mtoon';
 export * from '@pixiv/three-vrm-node-constraint';
 export * from '@pixiv/three-vrm-springbone';
+
+export type { VRMMaterialsV0CompatPlugin } from '@pixiv/three-vrm-materials-v0compat';
```
0b5vr, line 13: I think we can simply expose the entirety of `@pixiv/three-vrm-node-constraint` since it's already included in the build regardless.

yue4u:
> the report files end with `.api.md`, are those files meant to be committed to?

As post as  `Needs discussion:` in the description, with committed `api.md` CI will fail during the build if any unintended public api interface change detected.

cons: may be noisy / pollute search results

packages/three-vrm/src/VRMUtils/index.ts
0b5vr: mhm,,, How does it work? If I understand that the api-extractor simply forgets to output types for such static members, is that correct?

0b5vr:
> As post as `Needs discussion:` in the description, with committed `api.md` CI will fail during the build if any unintended public api interface change detected.

ooo that actually sounds sweet. While I slightly feel that I want to try this, by looking at other adopters no one is using this feature. Would you want to recommend us using the feature?

packages/three-vrm-core/src/expressions/index.ts
```diff
@@ -8,3 +8,6 @@ export { VRMExpressionMorphTargetBind } from './VRMExpressionMorphTargetBind';
 export { VRMExpressionOverrideType } from './VRMExpressionOverrideType';
 export { VRMExpressionPresetName } from './VRMExpressionPresetName';
 export { VRMExpressionTextureTransformBind } from './VRMExpressionTextureTransformBind';
+
+import type * as V0VRM from '@pixiv/types-vrm-0.0';
+export type { V0VRM };
```
0b5vr, line 13: If I understand correctly, these two lines are needed to control whether we are going to expose `V0VRM` from three-vrm-core or not, right?

Why we need to have this in `src/expressions/index.ts` , not `src/index.ts` ?

packages/three-vrm/src/VRMUtils/index.ts
yue4u: I added these to avoid api-extractor generating dts with some missing types. But after trying again now the issue has not been fixed. 🤔 Let me check it again.

yue4u: Ah, I think I finally figured this out.

1. `three-vrm` is exporting `VRMExpressionLoaderPlugin` form `three-vrm-core` which use `types-vrm-0.0` inside 
2. if we bundle `three-vrm-core` with `types-vrm-0.0` first, some exports from `types-vrm-0.0` became private in `three-vrm-core` but it works fine
4. if we bundle `three-vrm` with bundled `three-vrm-core` then, the private part is untouchable from `three-vrm` so the output dts is incomplete
    1. which seems reasonable so it's hard to say it's a bug in `api-extractor`

The same thing happends to `ConstraintSchema`.

So our options are

1. do not bundle dts from `types-*` packages in intermediate packages `three-vrm-*` and only bundle them inside  `three-vrm`
    1. can achieve single dts and no deps install
    2. some exports from `types-*` are still private so maybe hard to use
    3. types may conflict to user installed `types-*`
    4. introduce inconstancy
2. treat all `types-*` as dependencies or peerdependencies and let user install them
    1. not true dts bundle but flexible
    2. option to not install `types-vrm-0.0` at all
5. re-export or provide a public interface for each `types-*` in `three-vrm-*`
    1. can achieve single dts and no deps install
    3. needs dependency design?

What do you think?

0b5vr: Sounds like 2 is the best option...? It sounds easier from a contributor perspective at least.
3 is probably the most favorable option from a user perspective but it would make the module hard to maintain.

Issue #1294, opened 2023-09-22
Tracking issue: Revamp build process


yue4u:
Starting by https://github.com/pixiv/three-vrm/pull/1293 we are planned to revamp the build process.

Current TODOs:

- [ ] bundle dts as suggested in https://github.com/pixiv/three-vrm/issues/1289 https://github.com/pixiv/three-vrm/pull/1293
- [x] publish packages on CI https://github.com/pixiv/three-vrm/pull/1297
  - [x] adopt [npm package provenance](https://docs.npmjs.com/generating-provenance-statements)
- [x] check if we are still want to use rollup as the bundler, maybe swtich to tsup or esbuild #1411 
- [x] improve esm/cjs support #1411 
- [ ] maybe: mitigate to pnpm

PR #1297, opened 2023-09-26
feature: add publish action

yue4u:
https://github.com/pixiv/three-vrm/issues/1294

## Description

- add publish action via workflow_dispatch
- support npm provenance

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,68 @@
+name: Publish to npmjs
+
+on:
+  # Who has permission to workflow_dispatch
+  # https://github.com/orgs/community/discussions/26622
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to release'
+        required: true
+
+jobs:
+  fetch:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        # checkout is pinned to dev branch here
+        ref: dev
+    - name: Setup Node
+      uses: actions/setup-node@v3
+      with:
+        node-version-file: './.node-version'
+    - name: Cache Deps
+      uses: actions/cache@v3
+      with:
+        path: |
+          node_modules
+          packages/*/node_modules
+        key: ${{ runner.os }}-yarn-${{ hashFiles( 'yarn.lock' ) }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
+    - name: Fetch Deps
+      run: yarn install --frozen-lockfile
+
+  publish:
+    runs-on: ubuntu-latest
+    needs: fetch
+    permissions:
+      contents: write
+      id-token: write
+    steps:
+    - name: Set git username and email
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "actions-user"
+
+    - name: Set package version
+      run: |
+        yarn lerna version ${{ github.event.inputs.version }} --exact --no-push --yes
+        yarn lerna publish from-git --dist-tag latest
+      env:
+        NPM_CONFIG_PROVENANCE: true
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
+name: Publish to npmjs
+
+on:
+  # Who has permission to workflow_dispatch
+  # https://github.com/orgs/community/discussions/26622
+  workflow_dispatch:
```
yue4u, line 6: ` if: contains('["..."]', github.actor)` does not really make sense to me

.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
...
+  fetch:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        # checkout is pinned to dev branch here
+        ref: dev
+    - name: Setup Node
+      uses: actions/setup-node@v3
```
yue4u, line 28: v4 use node 20 we should update CI before that

.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
...
+          ${{ runner.os }}-yarn-
+    - name: Fetch Deps
+      run: yarn install --frozen-lockfile
+
+  publish:
+    runs-on: ubuntu-latest
+    needs: fetch
+    permissions:
+      contents: write
+      id-token: write
```
yue4u, line 48: https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions

.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+    - name: Set git username and email
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "actions-user"
+
+    - name: Set package version
+      run: |
+        yarn lerna version ${{ github.event.inputs.version }} --exact --no-push --yes
```
yue4u, line 57: I'm not sure if we need `--force-publish` here

.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
...
+        yarn lerna publish from-git --dist-tag latest
+      env:
+        NPM_CONFIG_PROVENANCE: true
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+    - name: Git push version up commits
+      run: |
+        git push origin ${{ github.ref_name }}
+
+    - name: Upload Builds
```
yue4u, line 67: Not sure if we want to automate the release tag on Github. There are only 3rd party solutions for now. https://github.com/actions/create-release#github-action---releases-api

yue4u (commented):
Anyway this might not work as intended unless we merge and test it on dev branch.
Current plan is to merge this and iterate through a few canary releases.

.github/workflows/publish.yml
0b5vr: `workflow_dispatch` is a slightly recent feature iirc; maybe they couldn't use the feature at that moment.
I agree with using `workflow_dispatch` :+1:

.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+    - name: Set git username and email
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "actions-user"
+
+    - name: Set package version
+      run: |
+        yarn lerna version ${{ github.event.inputs.version }} --exact --no-push --yes
+        yarn lerna publish from-git --dist-tag latest
```
0b5vr: it would be cool if we could make the disttag variable

yue4u: add in 42d5348c99f1c14872d500643622e91cad40b359

0b5vr:
let's try whether it's working properly

0b5vr (approved):

PR #1302, opened 2023-09-29
BREAKING: chore: remove downlevel-dts

yue4u:
BREAKING: TypeScript 3.9 is released in [May 2020](https://github.com/microsoft/TypeScript/releases/tag/v3.9.2). It does not make sense to support version before that.

1. Renamed all `.ts` -> `.d.ts` inside `types-*` packages
2. moved src -> types
3. remove building downlevel-dts but keeps files, ignores, etc

Work towards https://github.com/pixiv/three-vrm/pull/1293. This should also bring down build time a lot.


packages/three-vrm-core/.gitignore
```diff
@@ -0,0 +1 @@
+types/
```
yue4u: ~~Temp git ignore. After https://github.com/pixiv/three-vrm/pull/1293 types will end up in the lib dir~~ This is outdated

0b5vr:
@yue4u would you check `.eslintignore` and `.gitignore` ?
I searched the entire repo for `ts*` and there are these leftovers.

yue4u:
> @yue4u would you check `.eslintignore` and `.gitignore` ? I searched the entire repo for `ts*` and there are these leftovers.

a1167dd9ed2f385f8d18dd56218a0c5ae576a71b

0b5vr:
LGTM! thank you!

0b5vr (approved):

PR #1260, opened 2023-08-04
Bump three r154

ke456-png:
### Description

- 📦 Bumped Three.js to r154.
    - See: https://github.com/mrdoob/three.js/releases/tag/r154

PR #1312, opened 2023-10-11
fix: Fix pre-r154 incompatibility

0b5vr:
pre-r154 says `<colorspace_fragment>` is not defined

image

The change was introduced in https://github.com/pixiv/three-vrm/pull/1260, 8f0441aec569738e415291911705a92f36f3d5c1

Trivial: I just noticed that we still have several pre-r136 compats in our MToon implementation, what should we do...


ke456-png:
Thanks!

ke456-png (approved):

PR #1313, opened 2023-10-12
Publish workflow: Tags and docs

0b5vr:
- Revise CONTRIBUTING.md to use publish workflow
- Push git tags from publish workflow

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -59,8 +59,7 @@ jobs:
 
     - name: Git push version up commits
       run: |
-        git push origin ${{ github.ref_name }}
-        git push --tags
+        git push origin --tags ${{ github.ref_name }}
 
     - name: Upload Builds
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -60,6 +60,7 @@ jobs:
     - name: Git push version up commits
       run: |
         git push origin ${{ github.ref_name }}
+        git push --tags
```
yue4u:
```suggestion
        git push origin --tags ${{ github.ref_name }}
```

yue4u (approved):
Suggestion is optional. LGTM

.github/workflows/publish.yml
0b5vr: Didn't know this notation! looks good

See: https://git-scm.com/docs/git-push#Documentation/git-push.txt---tags

PR #1323, opened 2023-10-26
fix (three-vrm-core): Fix expressions material bind to respect alpha

### LINKED ISSUE #1301: Alpha Channel in Color Binds ignored
I have VRMs with expressions that change the alpha channel of a material's main color to make certain materials fade in and out.

The VRMExpressionMaterialColorBind's target value, as well as the MToon Material's color only store r,g,b. The alpha channel of the material is stored in .opacity on the material.
When parsing the VRM, the alpha channel of colorbinds needs to be added either as a new type of OpacityBind or on to the ColorBind

0b5vr:
will resolve #1301 

This PR fixes an issue that material binds of expression does not respect alpha values.

New optional property for the constructor of `VRMExpressionMaterialColorBind`: `targetAlpha`, which specifies the target alpha.

Alpha only applies to `"color"` type as spec specifies: https://github.com/vrm-c/vrm-specification/blob/master/specification/VRMC_vrm-1.0/expressions.md#materialcolorbind

Points need review:

- [ ] Is this working properly?
- [ ] Let me know if you spot any undesired breaking change

### DIFF
--- a/packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
+++ b/packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
@@ -73,8 +73,8 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
   public readonly targetAlpha: number;
 
   /**
-   * Its state.
-   * If it cannot find the target property in constructor, it will be null instead.
+   * Its binding state.
+   * If it cannot find the target property in the constructor, each property will be null instead.
    */
   private _state: BindState;
 
--- a/packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
+++ b/packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
@@ -30,8 +30,10 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
    * The first element stands for color channels, the second element stands for the alpha channel.
    * The second element can be null if the target property doesn't exist.
    */
+  // TODO: We might want to use the `satisfies` operator once we bump TS to 4.9 or higher
+  // See: https://github.com/pixiv/three-vrm/pull/1323#discussion_r1374020035
   private static _propertyNameMapMap: {
-    [distinguisher: string]: { [type in VRMExpressionMaterialColorType]?: [string, string | null] };
+    [distinguisher: string]: { [type in VRMExpressionMaterialColorType]?: readonly [string, string | null] };
   } = {
     isMeshStandardMaterial: {
       color: ['color', 'opacity'],
@@ -204,7 +206,9 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
     return { propertyName, initialValue, deltaValue };
   }
 
-  private _getPropertyNameMap(): { [type in VRMExpressionMaterialColorType]?: [string, string | null] } | null {
+  private _getPropertyNameMap():
+    | { [type in VRMExpressionMaterialColorType]?: readonly [string, string | null] }
+    | null {
     return (
       Object.entries(VRMExpressionMaterialColorBind._propertyNameMapMap).find(([distinguisher]) => {
         return (this.material as any)[distinguisher] === true;
--- a/packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
+++ b/packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
@@ -195,6 +195,16 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
     const propertyNameMap = this._getPropertyNameMap();
     const propertyName = propertyNameMap?.[type]?.[1] ?? null;
 
+    if (propertyName == null && targetAlpha !== 1.0) {
+      console.warn(
+        `Tried to add a material alpha bind to the material ${
+          material.name ?? '(no name)'
+        }, the type ${type} but the material or the type does not support alpha.`,
+      );
+
+      return null;
+    }
+
     if (propertyName == null) {
       return null;
     }
--- a/packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
+++ b/packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
@@ -4,30 +4,49 @@ import type { VRMExpressionMaterialColorType } from './VRMExpressionMaterialColo
 
 const _color = new THREE.Color();
 
+interface ColorBindState {
+  propertyName: string;
+  initialValue: THREE.Color;
+  deltaValue: THREE.Color;
+}
+
+interface AlphaBindState {
+  propertyName: string;
+  initialValue: number;
+  deltaValue: number;
+}
+
+interface BindState {
+  color: ColorBindState | null;
+  alpha: AlphaBindState | null;
+}
+
 /**
  * A bind of expression influences to a material color.
  */
 export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
   /**
    * Mapping of property names from VRMC/materialColorBinds.type to three.js/Material.
+   * The first element stands for color channels, the second element stands for the alpha channel.
+   * The second element can be null if the target property doesn't exist.
    */
   private static _propertyNameMapMap: {
-    [distinguisher: string]: { [type in VRMExpressionMaterialColorType]?: string };
+    [distinguisher: string]: { [type in VRMExpressionMaterialColorType]?: [string, string | null] };
   } = {
     isMeshStandardMaterial: {
-      color: 'color',
-      emissionColor: 'emissive',
+      color: ['color', 'opacity'],
+      emissionColor: ['emissive', null],
     },
     isMeshBasicMaterial: {
-      color: 'color',
+      color: ['color', 'opacity'],
     },
     isMToonMaterial: {
-      color: 'color',
-      emissionColor: 'emissive',
-      outlineColor: 'outlineColorFactor',
-      matcapColor: 'matcapFactor',
-      rimColor: 'parametricRimColorFactor',
-      shadeColor: 'shadeColorFactor',
+      color: ['color', 'opacity'],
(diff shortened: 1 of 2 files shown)


packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -75,84 +96,119 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
...
+    if (alpha != null) {
+      const { propertyName, initialValue } = alpha;
 
-    if (typeof (this.material as any).shouldApplyUniforms === 'boolean') {
-      (this.material as any).shouldApplyUniforms = true;
+      const target = (this.material as any)[propertyName] as number;
+      if (target != undefined) {
+        ((this.material as any)[propertyName] as number) = initialValue;
+      }
     }
```
0b5vr, line 159: set `MToonMaterial.shouldApplyUniforms` to `true` used to be required in past versions but now it's no longer required, so I removed the code section.

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -75,84 +96,119 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
      * The target color.
      */
     targetValue: THREE.Color;
+
+    /**
+     * The target alpha.
+     */
+    targetAlpha?: number;
```
0b5vr, line 105: this is the new constructor property

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -46,20 +65,22 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
    */
   public readonly targetValue: THREE.Color;
 
+  /**
+   * The target alpha.
+   */
+  public readonly targetAlpha: number;
```
0b5vr, line 73: this is the new public member

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -4,30 +4,49 @@ import type { VRMExpressionMaterialColorType } from './VRMExpressionMaterialColo
...
  */
 export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
   /**
    * Mapping of property names from VRMC/materialColorBinds.type to three.js/Material.
+   * The first element stands for color channels, the second element stands for the alpha channel.
+   * The second element can be null if the target property doesn't exist.
    */
   private static _propertyNameMapMap: {
-    [distinguisher: string]: { [type in VRMExpressionMaterialColorType]?: string };
+    [distinguisher: string]: { [type in VRMExpressionMaterialColorType]?: [string, string | null] };
```
yue4u: we might want to move to somthing like
```ts
private static _propertyNameMapMap: {
  // ...
} as const satisfies {
  [distinguisher: string]: { [type in VRMExpressionMaterialColorType]?: readonly [string, string | null] };
}
```
to reduce type casting in the future.

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -46,20 +65,22 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
...
   public readonly targetValue: THREE.Color;
 
+  /**
+   * The target alpha.
+   */
+  public readonly targetAlpha: number;
+
   /**
    * Its state.
    * If it cannot find the target property in constructor, it will be null instead.
```
yue4u: Should this comment be changed?

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -75,84 +96,119 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
...
+  }
 
-    if (propertyName == null) {
-      console.warn(
-        `Tried to add a material color bind to the material ${
-          material.name ?? '(no name)'
-        }, the type ${type} but the material or the type is not supported.`,
-      );
+  public applyWeight(weight: number): void {
+    const { color, alpha } = this._state;
```
yue4u: git diff is pretty chaos here in the diff viewer... had to checkout two versions and compare manually.

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -75,84 +96,119 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
...
 
-      this._state = null;
-    } else {
-      const target = (material as any)[propertyName] as THREE.Color;
+    if (color != null) {
+      const { propertyName, deltaValue } = color;
 
-      const initialValue = target.clone();
+      const target = (this.material as any)[propertyName] as THREE.Color;
```
yue4u, line 124: memo: wondering if change material  type to`THREE.MToonMaterial | THREE.MeshStandardMaterial | THREE.MeshBasicMaterial | THREE.Material` will improve types here

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -75,84 +96,119 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
...
   }
 
-  public clearAppliedWeight(): void {
-    if (this._state == null) {
-      // warning is already emitted in constructor
-      return;
+  private _initColorBindState(): ColorBindState | null {
+    const { material, type, targetValue } = this;
+
+    const propertyNameMap = this._getPropertyNameMap();
```
yue4u, line 165: we don't need to call `_getPropertyNameMap` twice separately in `_initColorBindState` and `_initAlphaBindState` since material will not change inside constructor? Anyway this won't be a trouble so optional.

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -75,84 +96,119 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
...
+    const { color, alpha } = this._state;
 
-    const { propertyName, deltaValue } = this._state;
+    if (color != null) {
+      const { propertyName, initialValue } = color;
 
-    const target = (this.material as any)[propertyName] as THREE.Color;
-    if (target === undefined) {
-      return;
-    } // TODO: we should kick this at `addMaterialValue`
```
yue4u, line 130: this TODO no longer relevant?

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
```diff
@@ -75,84 +96,119 @@ export class VRMExpressionMaterialColorBind implements VRMExpressionBind {
...
+    const { material, type, targetAlpha } = this;
 
-    target.copy(initialValue);
+    const propertyNameMap = this._getPropertyNameMap();
+    const propertyName = propertyNameMap?.[type]?.[1] ?? null;
 
-    if (typeof (this.material as any).shouldApplyUniforms === 'boolean') {
-      (this.material as any).shouldApplyUniforms = true;
+    if (propertyName == null) {
+      return null;
```
yue4u, line 209: should we warn for AlphaBind if not supported too?

yue4u (approved):
Didn't find any breaking change and sample works as intended. LGTM.

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
0b5vr: sounds good! adopted `readonly` for now.

https://github.com/pixiv/three-vrm/pull/1323/commits/8ccb147bca6d20dcb2c35101871cd94ba7140685

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
0b5vr: Classic git moment.
Basically I have moved some procedures in the constructor into its private methods (should've explained in advance 🙇 ).

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
0b5vr: Yes, it's kinda verbose. How do you think if we have a private member `_propertyNameMap` to cache the result?

0b5vr: mm, it's only for initialization, it doesn't affect the runtime performance. I think we're safe to leave it as is.

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
0b5vr: the `if (propertyName == null)` part in the `_initColorBindState` should already do the job if the end developer doesn't do the spicy thing to these materials.

packages/three-vrm-core/src/expressions/VRMExpressionMaterialColorBind.ts
0b5vr: No, we don't have to.
I wondered if we could check if the target alpha == 1 when the bind is for types that doesn't support alpha, just in case. How do you think?

yue4u: IMO it's a good to have but not necessary. It might be confusing when changing alpha for a wrong type but nothing happens.

0b5vr: Addressed @ https://github.com/pixiv/three-vrm/pull/1323/commits/2dded77dbe63ed3323f691c219ca92e6f72ca8be

PR #1332, opened 2023-11-22
Bump node to 20.10.0, Bump lint staged to 15.1.0

0b5vr:
This supercedes #1330 

This PR bumps lint-staged to 15.1.0.

lint-staged v15 requires node 18.2.0 at least, this PR also bumps node to 20.9.0 (LTS latest) as well.

See: https://github.com/lint-staged/lint-staged/releases/tag/v15.0.0

Also, the previous installation of lint-staged was not working correctly, making a lot of git stash and fails, sometimes deleting the staged / unstaged changes without commit.
This PR should resolve the behavior.


.node-version
```diff
@@ -1 +1 @@
-16.17.1
+20.9.0
```
yue4u: https://nodejs.org/en/blog/release/v20.10.0 (LTS) just released so maybe consider 20.10.0 too. Keep it as 20.9.0 is also ok.

packages/three-vrm-core/package.json
```diff
@@ -52,7 +52,6 @@
   },
   "devDependencies": {
     "@types/three": "^0.154.0",
-    "lint-staged": "13.1.2",
```
yue4u, line 55: Root lint-staged seems to work with per-package "lint-staged" config but currently html/js files aka `lint-examples` is not included. Is this intended?

.node-version
0b5vr: gonna bump to 20.10.0

packages/three-vrm-core/package.json
0b5vr: `lint-examples` should also be handled by lint-staged, indeed! Will do in a separated PR.

.node-version
0b5vr: addressed @ https://github.com/pixiv/three-vrm/pull/1332/commits/3a2e3a06e63e228abb7f285e011d3ffbbb82fc84

yue4u (approved):

PR #1342, opened 2023-12-04
fix: Fix the issue that hips translation goes infinity

0b5vr:
Fix the issue that hips translation goes infinity.

The `restHipsPosition` was always set to `Vector3(0, 0, 0)` which is pretty wrong.

I was probably quite sleepy when working on the restHipsPosition procedure last time

See: https://github.com/pixiv/three-vrm/commit/c11d6e9de74b1621292c5250cae7ae98cf26524a


ke456-png (approved):
Root motion now plays correctly. 

NICE! 👍

PR #1304, opened 2023-10-01
fix: Fix compatibility breaks at three r157 on MToon

MtBlue81:
## Problem
- Update to r157 in three.js causes mtoon shader error.
CleanShot 2023-10-01 at 14 30 38

## Changes
- Add compat code (
  See: https://github.com/mrdoob/three.js/pull/26805

PR #1348, opened 2024-01-10
deps: Bump three to r160

0b5vr:
This PR will bump Three.js to r160.

- Since it's bumped from r154, we have to address the change regarding the legacy light.
    - Updated all examples because of this change.
    - You have to multiply the directional light intensity by PI since now the light unit is set to a physically based one by default.
- I also added several COMPAT comments on pre-156 compats addressed in #1304.

I checked that all examples are working properly.

### DIFF
--- a/packages/three-vrm-materials-hdr-emissive-multiplier/examples/loader-plugin.html
+++ b/packages/three-vrm-materials-hdr-emissive-multiplier/examples/loader-plugin.html
@@ -59,7 +59,7 @@
 			const scene = new THREE.Scene();
 
 			// light
-			const light = new THREE.DirectionalLight( 0xffffff );
+			const light = new THREE.DirectionalLight( 0xffffff, Math.PI );
 			scene.add( light );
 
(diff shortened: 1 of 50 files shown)


0b5vr:
rebase ✅

packages/three-vrm-materials-hdr-emissive-multiplier/examples/loader-plugin.html
```diff
@@ -59,7 +59,7 @@
 			const scene = new THREE.Scene();
 
 			// light
-			const light = new THREE.DirectionalLight( 0xffffff );
+			const light = new THREE.DirectionalLight( 0xffffff, Math.PI );
```
ke456-png, line 62: ✅ See: https://github.com/mrdoob/three.js/pull/26392

ke456-png:
I checked examples are working in three.js r160 & r154

ke456-png (approved):
NICE! 👍

PR #1353, opened 2024-01-18
fix: Normalized bones now respect non humanoid parent rotations

0b5vr:
Normalized bones now respect non humanoid parent rotations.

When the humanoid bone hierarchy has non-humanoid bones with rotation (e.g., hips have a parent node with rotation), normalized bones don't work properly.
This commit will fix this.

### DIFF
--- a/packages/three-vrm-core/src/humanoid/VRMHumanoidRig.ts
+++ b/packages/three-vrm-core/src/humanoid/VRMHumanoidRig.ts
@@ -21,10 +21,11 @@ export class VRMHumanoidRig extends VRMRig {
     const root = new THREE.Object3D();
     root.name = 'VRMHumanoidRig';
 
-    // store boneWorldPositions and boneWorldRotations
+    // store boneWorldPositions, boneWorldRotations, and parentWorldRotations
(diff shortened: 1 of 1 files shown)


packages/three-vrm-core/src/humanoid/VRMHumanoidRig.ts
```diff
@@ -54,15 +57,13 @@ export class VRMHumanoidRig extends VRMRig {
...
-        let parentWorldRotation: THREE.Quaternion | undefined;
-        while (parentWorldPosition == null) {
+        let parentBoneWorldPosition: THREE.Vector3 | undefined;
+        while (parentBoneWorldPosition == null) {
           currentBoneName = VRMHumanBoneParentMap[currentBoneName];
           if (currentBoneName == null) {
             break;
           }
-          parentWorldPosition = boneWorldPositions[currentBoneName];
-          parentWorldRotation = boneWorldRotations[currentBoneName];
```
ke456-png, line 65: [memo]: VRMHumanBoneParentMap was used to reference a rotation of the parent bone, which was problematic when the parent was a non-humanoid bone with rotation.

ke456-png (approved):
Looks Good.

PR #1359, opened 2024-01-26
Remove `.needsUpdate` after texture transforms

### LINKED ISSUE #1358: [BUG] Significant performance bottleneck updating `VRMExpressionTextureTransformBind` textures
**Operating System**: MacOS
**Affected browsers**: Chrome (Metal backend) and Safari

_Note that this doesn't affect Chrome's OpenGL backend, only the Metal backend. For me, Metal recently became the default. You can choose a specific backend via chrome://flags#use-angle_

When calling `VRM.update()` on an asset that contains texture transform expression binds, I experience a 66-75% drop in FPS. Specifically, this happens during the `VRMExpressionManager.update()` call.

|Browser|With `VRMExpressionManager.update()`| Without `VRMExpressionManager.update()`|
|--|--|--|
|Chrome|CleanShot 2024-01-26 at 11 26 56@2x|CleanShot 2024-01-26 at 11 26 35@2x|
|Safari|CleanShot 2024-01-26 at 11 28 07@2x|CleanShot 2024-01-26 at 11 27 57@2x|

Chrome's performance profiler indicates that there is a significant bottleneck on the `texSubImage2D` call:

CleanShot 2024-01-26 at 11 32 06@2x

I narrowed the cause of this down to two lines in the `VRMExpressionTextureTransformBind` implementation where `.needsUpdate = true` is set on the transformed textures.

https://github.com/pixiv/three-vrm/blob/754f12dffe3c0bc946a063d1efe9d78183172e7a/packages/three-vrm-core/src/expressions/VRMExpressionTextureTransformBind.ts#L134

https://github.com/pixiv/three-vrm/blob/754f12dffe3c0bc946a063d1efe9d78183172e7a/packages/three-vrm-core/src/expressions/VRMExpressionTextureTransformBind.ts#L148

These `.needsUpdate` assignments are unnecessary because we are only modifying the `.repeat` and `.offset` vectors of the texture. Notice that this Three.js example freely updates those texture properties without flagging `.needsUpdate`: https://github.com/mrdoob/three.js/blob/02d4c5aa091ea8b6137db5bf1ad38819308246e3/examples/webgl_raycaster_texture.html#L338-L342

Removing these two lines resolves the performance bottleneck without affecting the texture transform behavior.

Demo: https://stackblitz.com/edit/github-otoerm?file=src%2Fmain.ts

https://github.com/pixiv/three-vrm/assets/22670878/47bf20a2-390c-450e-90b8-ff2d73220101

I suggest removing those two `.needsUpdate = true` lines. I can make a PR for this.

mattrossman:
Resolves #1358

### DIFF
--- a/packages/three-vrm-core/src/expressions/VRMExpressionTextureTransformBind.ts
+++ b/packages/three-vrm-core/src/expressions/VRMExpressionTextureTransformBind.ts
@@ -130,8 +130,6 @@ export class VRMExpressionTextureTransformBind implements VRMExpressionBind {
 
       target.offset.add(_v2.copy(property.deltaOffset).multiplyScalar(weight));
       target.repeat.add(_v2.copy(property.deltaScale).multiplyScalar(weight));
-
-      target.needsUpdate = true;
     });
   }
 
@@ -144,8 +142,6 @@ export class VRMExpressionTextureTransformBind implements VRMExpressionBind {
 
       target.offset.copy(property.initialOffset);
       target.repeat.copy(property.initialScale);
-
-      target.needsUpdate = true;
     });
   }
 }


0b5vr:
I see no clue why `target.needsUpdate = true` exists there.
r137 (our minimal version support) doesn't require `target.needsUpdate` at least.
Let's try removing these lines.

Thank you!

PR #1370, opened 2024-02-16
Include three-vrm-animation in three-vrm and export

ke456-png:
Include `three-vrm-animation` in `three-vrm`.


ke456-png:
Please import from `@pixiv/three-vrm-animation`.

PR #1374, opened 2024-02-26
fix: Fix type definition of VRM Animation

0b5vr:
Fixed type definition of `VRMC_vrm_animation`.

- Humanoid is optional
- Properties of HumanoidHumanBones are optional - Some bones are required but that would be handled in the implementation side - I also fixed the `VRMC_vrm` one

See: https://github.com/vrm-c/vrm-specification/blob/2d40293a5aa21f1a5011d7a53074056bd31cbad3/specification/VRMC_vrm_animation-1.0/schema/VRMC_vrm_animation.schema.json
See: https://github.com/vrm-c/vrm-specification/blob/2d40293a5aa21f1a5011d7a53074056bd31cbad3/specification/VRMC_vrm_animation-1.0/schema/VRMC_vrm_animation.humanoid.humanBones.schema.json


ke456-png:
[memo] specVersionだけが必須

ke456-png (approved):
nice !

PR #1376, opened 2024-02-26
fix: Fix internal dependency version from three-vrm-animation to three-vrm to latest

0b5vr:
- Fix dependency version to `@pixiv/three-vrm` in VRMA examples
- Fix dependency version from `@pixiv/three-vrm-animation` -> `@pixiv/types-vrmc-vrm-1.0` to latest


ke456-png (approved):
nice!

PR #1379, opened 2024-03-06
Bump Three.js to r162

0b5vr:
This PR bumps Three.js to r162.

See: https://github.com/mrdoob/three.js/releases/tag/r162

`@types/three` no longer expose THREE to global and I somehow forgot to import the types in many files, so I fixed it also.

### DIFF
--- a/packages/three-vrm-core/src/VRMCoreParameters.ts
+++ b/packages/three-vrm-core/src/VRMCoreParameters.ts
@@ -1,3 +1,4 @@
+import type * as THREE from 'three';
 import type { VRMExpressionManager } from './expressions/VRMExpressionManager';
 import type { VRMFirstPerson } from './firstPerson/VRMFirstPerson';
 import type { VRMHumanoid } from './humanoid/VRMHumanoid';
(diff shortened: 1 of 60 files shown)


yue4u:
Migration Guide

- https://github.com/mrdoob/three.js/wiki/Migration-Guide#r160--r161
- https://github.com/mrdoob/three.js/wiki/Migration-Guide#r161--r162

> WebGLRenderer now specifies precision qualifiers for all GLSL sampler types. https://github.com/mrdoob/three.js/pull/27482/files

This change does not affect mtoon and tests are also ok.

packages/three-vrm-core/src/VRMCoreParameters.ts
```diff
@@ -1,3 +1,4 @@
+import type * as THREE from 'three';
```
yue4u, line 1: This can also be done as `import { type Group } from 'three';` but requires more changes overall. I'm ok with current approach.

yue4u (approved):

PR #1384, opened 2024-03-25
BREAKING: WebGPU support (MToonNodeMaterial)

0b5vr:
This PR supports WebGPURenderer by adding `MToonNodeMaterial`, which is a NodeMaterial port of `MToonMaterial`.

After the change is published, you can use three-vrm on WebGPU by using `MToonNodeMaterialLoaderPlugin`.
To use the loader, specify the loader in the option of `VRMLoaderPlugin`.
See the `webgpu-dnd.html` example for details.

### BREAKING

- `gltf.userData.vrmMToonMaterials` might have materials other than `MToonMaterial` (namely, `MToonNodeMaterial` in this case).
    - This should not affect many users since the interface of `MToonNodeMaterial` is almost identical to `MToonMaterial`.

### TODOs

- [x] Since `three-vrm-material-mtoon` imports `three/addons/nodes/Nodes.js`, the initial module loading time became longer when you are using certain CDNs like unpkg. We are currently seeking for a workaround.
    - See also: https://discourse.threejs.org/t/nodematerial-on-unpkg-slow-to-load-all-the-dependencies/63213
    - Replaced the CDN from unpkg to jsdelivr and the problem is resolved.
        - See: https://github.com/mrdoob/three.js/pull/28006
        - See: https://github.com/pixiv/three-vrm/pull/1385
- [x] Several type definitions on `@types/three` are patched by `patch-package`
    - We might want to fix by `three-ts-types` side before we ship this
- [x] The `Type of property 'cache' circularly references itself ...` error (TS2615) occurs
    - We might want to bump our TypeScript version to 5.1 or later
    - The issue is resolved, at least in the patch
        - See: https://github.com/three-types/three-ts-types/pull/744#issuecomment-2036440655


0b5vr:
Examples are working properly with [r163](https://github.com/mrdoob/three.js/releases/tag/r163).

0b5vr:
Ready to merge!

We are going to bump the major since the breaking change occurs.

0b5vr:
日本語で恐縮ですが、NodeMaterialを継承したマテリアルのコードを読む上で、NodeMaterialの動作の流れ（r164現在のもの）を以下のページにまとめてあります。レビューの際にご活用ください。

https://scrapbox.io/0b5vr/Three.js:_NodeMaterial

packages/three-vrm-materials-mtoon/src/nodes/MToonNodeMaterialLoaderPlugin.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export class MToonNodeMaterialLoaderPlugin extends MToonMaterialLoaderPlugin {
+  public constructor(parser: GLTFParser, options: MToonMaterialLoaderPluginOptions = {}) {
+    super(parser, options);
+  }
+
+  public getMaterialType(materialIndex: number): typeof THREE.Material | null {
+    const v1Extension = this._getMToonExtension(materialIndex);
+    if (v1Extension) {
+      return MToonNodeMaterial;
```
ke456-png, line 15: [memo]: return MToonNodeMaterial

packages/three-vrm-materials-mtoon/src/MToonMaterialLoaderPlugin.ts
```diff
@@ -146,7 +146,7 @@ export class MToonMaterialLoaderPlugin implements GLTFLoaderPlugin {
     });
   }
 
-  private _getMToonExtension(materialIndex: number): V1MToonSchema.VRMCMaterialsMToon | undefined {
+  protected _getMToonExtension(materialIndex: number): V1MToonSchema.VRMCMaterialsMToon | undefined {
```
ke456-png, line 149: [memo] It is necessary to be referenced from MToonNodeMaterialLoaderPlugin.getMaterialType()

ke456-png:
The rendering result differs when there is no texture and only a normal map.

examples/feature-test
image

webgpu-feature-test
image

ke456-png:
Perhaps MToonMaterial was wrong?


https://github.com/pixiv/three-vrm/blob/8a8944949e6475e1a418a260dcf443a6f6764289/packages/three-vrm-materials-mtoon/src/MToonMaterial.ts#L607-L614

```
 const useUvInFrag = 
...
      this.uvAnimationMaskTexture !== null ||
      this.normalMap !== null;

```

0b5vr:
> The rendering result differs when there is no texture and only a normal map.

It seems to be, yes! The blue ball in the example should have bumps on the surface since the normal map is assigned.
I will fix this in another PR.

ke456-png:
If possible, please incorporate this change.
https://github.com/pixiv/three-vrm/commit/25cc2ba3f40c0f80d7da4a51a8e4cd84dc0a79ab

packages/three-vrm-materials-mtoon/src/nodes/MToonNodeMaterial.ts
```diff
@@ -0,0 +1,485 @@
...
+      this.vertexColors = false;
+    }
+
+    // the ordinary diffuseColor setup
+    super.setupDiffuseColor(builder);
+
+    // Set alpha to 1 if it is not transparent
+    if (this.transparent === false) {
+      Nodes.diffuseColor.a.assign(1.0);
+    }
```
ke456-png: I need to confirm the conditions for making it opaque.

https://github.com/pixiv/three-vrm/blob/9e16c3855a5b441cca0f11480c4003de25d2943c/packages/three-vrm-materials-mtoon/src/shaders/mtoon.frag#L871-L873

https://github.com/mrdoob/three.js/blob/687a4caf8c5d633df691a90369c9c185a4acb0fc/src/renderers/webgl/WebGLPrograms.js#L250
`opaque: material.transparent === false && material.blending === NormalBlending && material.alphaToCoverage === false,`

0b5vr:
Resolved conflict with `dev-v3`, it's now ready to build using esbuild.
I don't have to explicitly specify `three/addons/nodes/Nodes.js`, esbuild seems to handle this by only specifying `three`.

0b5vr:
> If possible, please incorporate this change. [25cc2ba](https://github.com/pixiv/three-vrm/commit/25cc2ba3f40c0f80d7da4a51a8e4cd84dc0a79ab)

yes I have to indeed

0b5vr:
> > If possible, please incorporate this change. [25cc2ba](https://github.com/pixiv/three-vrm/commit/25cc2ba3f40c0f80d7da4a51a8e4cd84dc0a79ab)
> 
> yes I have to indeed

done @ https://github.com/pixiv/three-vrm/pull/1384/commits/36f064fb7653920ec4ff3bf453ac478ba08432dc

packages/three-vrm-materials-mtoon/src/nodes/MToonNodeMaterial.ts
0b5vr: changed that to follow the WebGLRenderer way for now.

https://github.com/pixiv/three-vrm/pull/1384/commits/afe9143d2054097c1b99aae876a77d4bbe58c38c

0b5vr:
`MToonNodeMaterial` is gonna be compatible only with r161 or higher.
I added a warning if the user attempts to use this in r160 or before.

https://github.com/pixiv/three-vrm/pull/1384/commits/afe9143d2054097c1b99aae876a77d4bbe58c38c

packages/three-vrm-materials-mtoon/src/nodes/MToonNodeMaterial.ts
ke456-png: NICE!

ke456-png (approved):
Looks good.

PR #1385, opened 2024-03-27
chore: change cdn from unpkg to jsDelivr

0b5vr:
This PR replaces CDN from unpkg to jsDelivr.
I confirmed that all examples are working properly on my end.

Also update README to recommend jsDelivr.

Following the Three.js side change.
See: https://github.com/mrdoob/three.js/pull/28006

es-module-shims for import map is also no longer used in Three.js examples. See: https://github.com/mrdoob/three.js/pull/26836


ke456-png (approved):
NICE! 👍

PR #1395, opened 2024-05-05
feat: add experimental @pixiv/vrm-viewer package

yue4u:
Add experimental `@pixiv/vrm-viewer` package via [model-viewer](https://modelviewer.dev/)

### DIFF
--- a/bin/build.mjs
+++ b/bin/build.mjs
@@ -8,6 +8,7 @@ const NODE_ENV = process.env.NODE_ENV;
 const DEV = NODE_ENV === 'development';
 const SERVE = process.env.SERVE === '1';
 const PORT = parseInt(process.env.PORT || '10001', 10);
+const minifySuffix = DEV ? '' : '.min';
 
 // == build =========================================================================================
 
@@ -17,12 +18,16 @@ await buildPackage(workDir);
 // == implement =====================================================================================
 
 /**
+ * @typedef {{ name: string, version: string, description: string, module: string }} PackageJson
+ */
+
+/**
  *
  * @param {string} absWorkingDir
  */
 async function buildPackage(absWorkingDir) {
   /**
-   * @type {{ name: string, version: string, description: string, module: string }}
+   * @type {PackageJson}
    */
   const packageJson = JSON.parse(await fs.readFile(path.join(absWorkingDir, './package.json'), 'utf-8'));
 
@@ -55,8 +60,8 @@ async function buildPackage(absWorkingDir) {
    * @param {'esm' | 'cjs'} format
    */
   const entryPoints = (format) => {
-    let outFilename = filename + (format === 'esm' ? '.module' : '');
-    outFilename += DEV ? '' : '.min';
+    const esmSuffix = format === 'esm' ? '.module' : '';
+    const outFilename = filename + esmSuffix + minifySuffix;
     return {
       entryPoints: {
         [outFilename]: 'src/index.ts',
@@ -78,6 +83,8 @@ async function buildPackage(absWorkingDir) {
     loader: { '.frag': 'text', '.vert': 'text' },
     sourcemap: DEV ? 'inline' : false,
     absWorkingDir,
+    outdir: 'lib',
+    tsconfig: path.join(absWorkingDir, 'tsconfig.json'),
     logLevel: 'info',
     color: true,
     minify: !DEV,
@@ -89,11 +96,15 @@ async function buildPackage(absWorkingDir) {
     const esmContext = await esbuild.context({
       ...buildOptions,
       ...entryPoints('esm'),
-      outdir: 'lib',
       absWorkingDir,
     });
 
-    await esmContext.watch();
+    const extraContext =
+      packageJson.name === '@pixiv/vrm-viewer'
+        ? (await esbuild.context(vrmViewOptions(buildOptions))).watch()
+        : Promise.resolve();
+
+    await Promise.all([esmContext.watch(), extraContext]);
     await esmContext.serve({ servedir: '.', port: PORT });
   }
 
@@ -101,20 +112,37 @@ async function buildPackage(absWorkingDir) {
     await esbuild.build({
       ...buildOptions,
       ...entryPoints('esm'),
-      outdir: 'lib',
     });
 
     await esbuild.build({
       ...buildOptions,
       ...entryPoints('cjs'),
-      outdir: 'lib',
       outExtension: {
         // eslint-disable-next-line @typescript-eslint/naming-convention
         '.js': '.cjs',
       },
     });
+
+    if (packageJson.name === '@pixiv/vrm-viewer') {
+      await esbuild.build(vrmViewOptions(buildOptions));
+    }
   }
 
   // == main =========================================================================================
   await (SERVE ? serve : build)();
+}
+
+/**
+ * @param {import('esbuild').BuildOptions} baseBuildOptions
+ * @returns {import('esbuild').BuildOptions}
+ */
+function vrmViewOptions(baseBuildOptions) {
+  return {
+    ...baseBuildOptions,
+    format: 'esm',
+    entryPoints: {
+      [`vrm-viewer${minifySuffix}`]: 'src/index.ts',
+    },
+    external: [], // include three
+  };
 }
--- a/bin/build.mjs
+++ b/bin/build.mjs
@@ -135,6 +135,9 @@ async function buildPackage(absWorkingDir) {
 /**
  * @param {import('esbuild').BuildOptions} baseBuildOptions
  * @returns {import('esbuild').BuildOptions}
+ *
+ * NOTE: The main reason to separated this from the main config is that
+ * we want to include Three.js in the build but `external` is fixed per esbuild context
  */
 function vrmViewOptions(baseBuildOptions) {
   return {
--- a/package.json
+++ b/package.json
@@ -1,15 +1,9 @@
 {
   "private": true,
   "type": "module",
-  "workspaces": {
-    "packages": [
-      "packages/*"
-    ],
-    "nohoist": [
-      "**/@google/model-viewer",
-      "**/@google/model-viewer/**"
-    ]
-  },
+  "workspaces": [
+    "packages/*"
+  ],
   "publishConfig": {
     "provenance": true
   },
--- a/package.json
+++ b/package.json
@@ -1,9 +1,15 @@
 {
   "private": true,
   "type": "module",
-  "workspaces": [
-    "packages/*"
-  ],
+  "workspaces": {
+    "packages": [
+      "packages/*"
+    ],
+    "nohoist": [
+      "**/@google/model-viewer",
+      "**/@google/model-viewer/**"
+    ]
+  },
   "publishConfig": {
     "provenance": true
   },
--- a/packages/vrm-viewer/examples/example.html
+++ b/packages/vrm-viewer/examples/example.html
@@ -24,7 +24,8 @@
 </head>
 
 <body>
-	<vrm-viewer camera-controls touch-action="pan-y" src="./models/VRM1_Constraint_Twist_Sample.vrm" />
+	<vrm-viewer camera-controls touch-action="pan-y" tone-mapping="neutral"
+		src="./models/VRM1_Constraint_Twist_Sample.vrm" />
 	<script type="importmap">
 		{
 			"imports": {
--- a/packages/vrm-viewer/examples/example.html
+++ b/packages/vrm-viewer/examples/example.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+
+<html>
+
+<head>
+	<meta charset="utf-8" />
+	<title>three-vrm example</title>
+	<meta name="viewport"
+		content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no" />
+	<style>
+		body {
+			margin: 0;
+		}
+
+		canvas {
+			display: block;
+		}
+
+		vrm-viewer {
+			width: 100vw;
+			height: 100vw;
+		}
+	</style>
+</head>
+
+<body>
+	<vrm-viewer camera-controls touch-action="pan-y" src="./models/VRM1_Constraint_Twist_Sample.vrm" />
+	<script type="importmap">
+		{
+			"imports": {
+				"three": "https://cdn.jsdelivr.net/npm/three@0.162.0/build/three.module.js"
+			}
+		}
+	</script>
+	<script type="module" src="../lib/vrm-viewer.module.js"></script>
+</body>
+
+</html>
--- a/packages/vrm-viewer/examples/sperate-three.html
+++ b/packages/vrm-viewer/examples/sperate-three.html
@@ -29,8 +29,9 @@
 	<script type="importmap">
 		{
 			"imports": {
-				"three": "https://cdn.jsdelivr.net/npm/three@0.164.0/build/three.module.js",
-				"three/": "https://cdn.jsdelivr.net/npm/three@0.164.0/"
+				"three": "https://cdn.jsdelivr.net/npm/three@0.164.1/build/three.module.js",
+				"three/": "https://cdn.jsdelivr.net/npm/three@0.164.1/",
+				"three/addons/": "https://cdn.jsdelivr.net/npm/three@0.164.1/examples/jsm/"
 			}
 		}
 	</script>
--- a/packages/vrm-viewer/examples/sperate-three.html
+++ b/packages/vrm-viewer/examples/sperate-three.html
@@ -35,6 +35,37 @@
 		}
 	</script>
 	<script type="module" src="../lib/vrm-viewer.module.js"></script>
+	<script type="module">
+		import * as THREE from 'three';
+		const vrmViewer = document.querySelector( 'vrm-viewer' );
+
+		// https://modelviewer.dev/docs/index.html#loading-events
+
+		vrmViewer.addEventListener( 'model-visibility', ( event ) => {
+
+			console.log( `model visibility: ${event.detail.visible}` );
+
+		} );
+
+		vrmViewer.addEventListener( 'before-render', ( _event ) => {
+
+			const mesh = new THREE.Mesh(
+				new THREE.SphereGeometry( 0.1 ),
+				new THREE.MeshStandardMaterial( { color: 0xff8800 } ),
+			);
+			mesh.position.set( 0.5, 1.5, 0 );
+
+			vrmViewer.vrm.scene.add( mesh );
+
+		} );
+
+		vrmViewer.addEventListener( 'error', ( event ) => {
+
+			console.error( 'load error', event );
+
+		} );
+
+	</script>
 </body>
 
 </html>
--- a/packages/vrm-viewer/examples/sperate-three.html
+++ b/packages/vrm-viewer/examples/sperate-three.html
@@ -26,7 +26,15 @@
 <body>
 	<vrm-viewer camera-controls touch-action="pan-y" tone-mapping="neutral"
 		src="./models/VRM1_Constraint_Twist_Sample.vrm" />
-	<script type="module" src="../lib/vrm-viewer.js"></script>
+	<script type="importmap">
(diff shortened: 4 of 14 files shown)


packages/vrm-viewer/examples/example.html
```diff
@@ -0,0 +1,38 @@
...
+<body>
+	<vrm-viewer camera-controls touch-action="pan-y" src="./models/VRM1_Constraint_Twist_Sample.vrm" />
+	<script type="importmap">
+		{
+			"imports": {
+				"three": "https://cdn.jsdelivr.net/npm/three@0.162.0/build/three.module.js"
+			}
+		}
+	</script>
+	<script type="module" src="../lib/vrm-viewer.module.js"></script>
```
yue4u: It's possible to bundle three as a single `vrm-viewer.bundle.js` but due to some 
 current rollup version's limitations? mutiple three are bundled so kept it outside. https://modelviewer.dev/docs/faq.html#:~:text=How%20do%20I%20avoid%20duplicating%20three.js%20in%20my%20bundle%3F

packages/vrm-viewer/examples/models/VRM1_Constraint_Twist_Sample.vrm
yue4u, line 1: I think we should merge all examples folders at root dir to avoid duplicates of vrms, envmaps, etc...

packages/vrm-viewer/rollup.config.js
```diff
@@ -0,0 +1,101 @@
+/* eslint-env node */
```
yue4u: We should also avoid duplicating rollup config everywhere or swtich to something like esbuild when possible. Currently building the whole workspace takes several minutes on my machine and makes no sense.

packages/vrm-viewer/src/features/vrm-mixin.ts
```diff
@@ -0,0 +1,106 @@
+import { VRM, VRMLoaderPlugin, VRMUtils } from '@pixiv/three-vrm';
+import ModelViewerElementBase, {
```
yue4u, line 2: For reviewing this reading some `<model-viewer />`'s source code would be helpful.

packages/vrm-viewer/src/vrm-viewer.ts
```diff
@@ -0,0 +1,14 @@
+import { VRMMixin } from './features/vrm-mixin';
+import { ModelViewerElement } from '@google/model-viewer/src/model-viewer';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const VRMViewerElement = VRMMixin(ModelViewerElement);
```
yue4u: not work with [`<model-viewer-effects />`](https://modelviewer.dev/examples/postprocessing/index.html) yet due to node name check. I think it have to be fixed upstream.

packages/vrm-viewer/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@pixiv/vrm-viewer",
+  "version": "2.1.1",
+  "description": "VRM file loader for three.js.",
```
0b5vr:
```suggestion
  "description": "The <vrm-viewer> web component that displays an interactive VRM model viewer",
```

packages/vrm-viewer/examples/example.html
0b5vr: memo:

- The public model-viewer build includes three.js in the product
- rollup is also used in both Three.js and model-viewer

package.json
```diff
@@ -1,8 +1,14 @@
...
-    "packages/*"
-  ],
+  "workspaces": {
+    "packages": [
+      "packages/*"
+    ],
+    "nohoist": [
+      "**/@google/model-viewer",
+      "**/@google/model-viewer/**"
+    ]
```
0b5vr: what is the reason for these nohoist?

yue4u: `@rollup/plugin-node-resolve` only looks for node_modules inside build dir, so it's necessary to put related packages in the same dir that has package.json, aka avoid hoisting

packages/vrm-viewer/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["./src"],
+  "compilerOptions": {
+    "module": "ESNext",
+    "experimentalDecorators": true
```
0b5vr, line 6: What is this `experimentalDecorators`? is this required to compile the project with the `@google/model-viewer` dependency?

yue4u: Not used directly for now but required for building types for `@google/model-viewer`.   It's used for [extending web components's properties]( https://github.com/google/model-viewer/blob/master/packages/model-viewer/src/features/animation.ts#L52-L56).  We may want to add `vrma="..."` via `@property` too later.

package.json
0b5vr: Sorry but I couldn't understand the explanation. Could you add details or reword?

yue4u:
```sh
.
├── node_modules # root
│   ├── @google/model-viewer/...  # "hoist" to root
│   └── ...
├── package.json # root
├── packages
│   ├── vrm-viewer # 💥 can not resolve `@google/model-viewer` 
│   │   ├── package.json # vrm-viewer's build base dir
│   │   ├── src
│   │   └── types
└── yarn.lock
```


```sh
.
├── node_modules # root
│   └── ...
├── package.json # root 
├── packages
│   ├── vrm-viewer  # ✅ build ok
│   │   ├── node_modules # node_modules appears here by "nohoist"
│   │   │   └── @google/model-viewer/...
│   │   ├── package.json # vrm-viewer's build base dir
│   │   ├── src
│   │   └── types
└── yarn.lock
```

0b5vr: I see, thank you for the explanation!

packages/vrm-viewer/rollup.config.js
```diff
@@ -0,0 +1,101 @@
...
+import packageJson from './package.json';
+import serve from 'rollup-plugin-serve';
+import { string } from 'rollup-plugin-string';
+import { terser } from 'rollup-plugin-terser';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import typescript from '@rollup/plugin-typescript';
+
+// == constants ====================================================================================
+/** copyright text */
+const copyright = '(c) 2019-2024 pixiv Inc.';
```
0b5vr:
```suggestion
const copyright = '(c) 2024 pixiv Inc.';
```

packages/vrm-viewer/package.json
```diff
@@ -0,0 +1,67 @@
...
+    ]
+  },
+  "dependencies": {
+    "@google/model-viewer": "3.5.0",
+    "@pixiv/three-vrm": "2.1.1"
+  },
+  "devDependencies": {
+    "@rollup/plugin-node-resolve": "^15.0.1",
+    "@types/three": "^0.162.0",
+    "lit": "^3.1.3",
```
0b5vr, line 60: Where is `lit` used?

yue4u: model-viewer itself is built with `lit` https://github.com/google/model-viewer/blob/master/packages/model-viewer/package.json#L86 
usage in following PR: https://github.com/pixiv/three-vrm/pull/1398/files#diff-9afb2720e7e69f4a87c397651ebd81ebe30db7ed81590101fe225285b6a600b0R4

packages/vrm-viewer/rollup.config.js
```diff
@@ -0,0 +1,101 @@
...
+const copyright = '(c) 2024 pixiv Inc.';
+
+/** name of the license */
+const licenseName = 'MIT License';
+
+/** url of the license */
+const licenseUri = 'https://github.com/pixiv/three-vrm/blob/release/LICENSE';
+
+/** output name of the module */
+const globalName = 'THREE_VRM';
```
0b5vr:
```suggestion
const globalName = 'VRM_VIEWER';
```

bin/build.mjs
```diff
@@ -78,6 +83,8 @@ async function buildPackage(absWorkingDir) {
     loader: { '.frag': 'text', '.vert': 'text' },
     sourcemap: DEV ? 'inline' : false,
     absWorkingDir,
+    outdir: 'lib',
+    tsconfig: path.join(absWorkingDir, 'tsconfig.json'),
```
yue4u, line 87: needed for decorators

bin/build.mjs
```diff
@@ -89,32 +96,53 @@ async function buildPackage(absWorkingDir) {
     const esmContext = await esbuild.context({
       ...buildOptions,
       ...entryPoints('esm'),
-      outdir: 'lib',
       absWorkingDir,
     });
 
-    await esmContext.watch();
+    const extraContext =
```
yue4u, line 102: esbuild is clever enough to not bundle mutiple three, so created a bundle-all build

yue4u (commented):
Currently type build fails and waiting for https://github.com/google/model-viewer/issues/4786

https://github.com/pixiv/three-vrm/actions/runs/9203519761/job/25315221048#step:5:252

package.json
yue4u: This change no longer needs for esbuild so reverted.

bin/build.mjs
```diff
@@ -89,32 +96,53 @@ async function buildPackage(absWorkingDir) {
...
+function vrmViewOptions(baseBuildOptions) {
+  return {
+    ...baseBuildOptions,
+    format: 'esm',
+    entryPoints: {
+      [`vrm-viewer${minifySuffix}`]: 'src/index.ts',
+    },
+    external: [], // include three
+  };
+}
```
0b5vr, line 151: Please clarify the intention why we have to separate options only for vrm-viewer.

I guess it's because:

- we don't have to output `cjs` builds
- we want to specify that the output filename is `vrm-viewer.js` or `vrm-viewer.min.js`
- we want to include Three.js in the build

bin/build.mjs
0b5vr: > so created a bundle-all build

wdym

bin/build.mjs
yue4u: > we don't have to output cjs builds

YES, but can be avoided by reusing the esm context.

> we want to specify that the output filename is vrm-viewer.js or vrm-viewer.min.js

YES, but can be avoided by reusing existing context.

> we want to include Three.js in the build

YES, this is the main reason since `external` is fixed per context.

0b5vr:
`example.html` and `separate-three.html`, Do they mean that I should choose one of them?

yue4u:
> `example.html` and `separate-three.html`, Do they mean that I should choose one of them?

Those have different concerns and controls.
1. `example.html` for easiest vrm display, only using three-vrm + modal-viewer features, not assuming extending three/three-vrm via another build system.
2. `separate-three.html` for the possibility to use more three-related libraries, and the `.module` build can be consumed via any build system again. (I'm not very confident about the filename.)

bin/build.mjs
yue4u: When using rollup there is a issue that three is bundled multiple times and that issue is lifted after switching to esbuild; therefore It becomes possible to create a build that includes three (not treat three as external to avoid importmap at all).

bin/build.mjs
0b5vr: would you put them in the code as comments?

bin/build.mjs
0b5vr: What exactly is "bundle-all" here? Are you simply mentioning the `vrm-viewer.js` build includes Three.js?

bin/build.mjs
0b5vr: what happens if we don't specify `tsconfig` here? Does it attempt to discover `tsconfig.json` near from its project root instead of the cwd?

packages/vrm-viewer/rollup.config.js
0b5vr: Addressed @ #1411 🎉

packages/vrm-viewer/examples/example.html
0b5vr: now we switched to esbuild

bin/build.mjs
yue4u: > Are you simply mentioning the vrm-viewer.js build includes Three.js

yes, including three,three-vrm,model-viewer

bin/build.mjs
yue4u: https://github.com/pixiv/three-vrm/pull/1395/commits/76a31409fb74b22791cc96cf7b30a3e454d87607

packages/vrm-viewer/types-shim/index.d.ts
```diff
@@ -0,0 +1,36 @@
+// we should remove this once upstream supports
```
yue4u: generated with `build-types`

bin/build.mjs
0b5vr: so now what is the purpose we provide the `separate-three.html` example?

packages/vrm-viewer/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "LICENSE"
+  ],
+  "type": "module",
+  "main": "lib/vrm-viewer.module.js",
+  "types": "types/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./types-shim/index.d.ts",
+      "import": "./lib/vrm-viewer.module.min.js",
+      "require": "./lib/vrm-viewer.min.cjs"
```
0b5vr:
```suggestion
      "import": "./lib/vrm-viewer.module.js",
      "require": "./lib/vrm-viewer.cjs"
```

See: https://github.com/pixiv/three-vrm/pull/1416

bin/build.mjs
0b5vr: This one right? If so I would want to see the actual use cases that combines existing Three.js libraries

https://github.com/pixiv/three-vrm/pull/1395#issuecomment-2128453450

yue4u: I don't think extra example is really needed, this example makes most sense by demonstrating it's possible to bring users' own three.js or have a build setup while avoiding duplication in dist bundle.

0b5vr: Such an example doesn't make any sense to most users if the purpose is not provided because it looks the same in appearance.

yue4u: To be honest the build is more intended to use with a build setup and we are limited to raw html here. Using Import map itself is a simulation but not a real world usage. So it's more like a test rather than example?

packages/vrm-viewer/types-shim/index.d.ts
0b5vr: wdym. does this get complained by eslint?

yue4u: https://github.com/pixiv/three-vrm/pull/1395#pullrequestreview-2073002062
1. model-viewer was not compliant with three v164 in `USDZExporter`, fix has been merged by not released
2. if we run tsc it can generate dts file but raise error
3. as experimental we do not care this error in model-viewer and `USDZExporter` at this point so commit types as shim
4. eslint is not related.

bin/build.mjs
0b5vr: Hmm. Yes, I kinda got the intention. Yet still feels like it will cause confusion unless we provide a reason somewhere.

0b5vr: We can provide an example that loads an extra model in the scene for example.

```js

const viewer = document.getElementById('viewer');

// See: https://modelviewer.dev/docs/faq.html#entrydocs-general-questions-three
const scene = viewer[Object.getOwnPropertySymbols(viewer).find(e => e.description === 'scene')];

const mesh = new THREE.Mesh(
	new THREE.SphereGeometry(0.1),
	new THREE.MeshStandardMaterial({ color: 0xff8800 }),
);
mesh.position.set(0.5, 0.7, 0);
scene.add(mesh);
```

image

packages/vrm-viewer/types-shim/index.d.ts
0b5vr: you are talking about `types-shim/features/vrm-mixin.d.ts:14` right?

https://github.com/pixiv/three-vrm/pull/1395/files#diff-0916a4a0925b42342f83992b5547feb1301d1ad2e3e7313ed3c572a44ed6891eR13

bin/build.mjs
yue4u: Made some changes in 507a8f3ff6f4f641c1c7de3a3ea1712312b580cd , what do you think?

packages/vrm-viewer/types-shim/index.d.ts
yue4u: No that's irrelevant. Just run `build-types` and you'll get
```
$ tsc --declaration --declarationDir ./types --emitDeclarationOnly
../../node_modules/@google/model-viewer/src/features/ar.ts:449:42 - error TS2554: Expected 3-4 arguments, but got 1.

449       const arraybuffer = await exporter.parse(model);
                                             ~~~~~

  ../../node_modules/@types/three/examples/jsm/exporters/USDZExporter.d.ts:13:9
    13         onDone: (result: Uint8Array) => void,
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'onDone' was not provided.

../../node_modules/@google/model-viewer/src/features/ar.ts:454:30 - error TS2322: Type 'void' is not assignable to type 'BlobPart'.

454       const blob = new Blob([arraybuffer], {
                                 ~~~~~~~~~~~


Found 2 errors in the same file, starting at: ../../node_modules/@google/model-viewer/src/features/ar.ts:449

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

packages/vrm-viewer/examples/sperate-three.html
```diff
@@ -0,0 +1,71 @@
...
+</head>
+
+<body>
+	<vrm-viewer camera-controls touch-action="pan-y" tone-mapping="neutral"
+		src="./models/VRM1_Constraint_Twist_Sample.vrm" />
+	<script type="importmap">
+		{
+			"imports": {
+				"three": "https://cdn.jsdelivr.net/npm/three@0.164.0/build/three.module.js",
+				"three/": "https://cdn.jsdelivr.net/npm/three@0.164.0/"
```
0b5vr:
```suggestion
				"three": "https://cdn.jsdelivr.net/npm/three@0.164.1/build/three.module.js",
				"three/": "https://cdn.jsdelivr.net/npm/three@0.164.1/",
				"three/addons/": "https://cdn.jsdelivr.net/npm/three@0.164.1/examples/jsm/"
```

The new build with WebGPU support needs `three/addons/` to be resolved.

packages/vrm-viewer/types-shim/index.d.ts
0b5vr: Thanks. Would you put [a link to the comment above](https://github.com/pixiv/three-vrm/pull/1395#discussion_r1621561490) as a comment?

yue4u: https://github.com/pixiv/three-vrm/pull/1395/commits/5f72879101b41d28aeb2058e912ff5ea24990ac0

bin/build.mjs
0b5vr: oh wait, you make `vrm` accessible from `vrmViewer`? That sounds weird a bit considering the original `model-viewer` design doesn't allow Three.js related members to be exposed

yue4u: I think It's necessary anyway and common when using `lit`. if one want to extend vrm-viewer and call three-vrm specific methods, doing this via symbols creates more burden than being helpful. Somehow similar to https://github.com/google/model-viewer/blob/master/packages/model-viewer/src/features/scene-graph.ts#L89-L91 ?
I have no strong ideas on this so changing to symbols is also ok to me.

0b5vr: Yes I agree that changing this to a symbol doesn't make any profit to both end developers and maintainers. However I want to know the intention why `model-viewer` makes such a design decision.

yue4u:
After discussion this and https://github.com/pixiv/three-vrm/pull/1398 will not be a part of three-vrm for now and likely to be publish under a separated namespace.

PR #1398, opened 2024-05-13
feat(vrm-viewer): experimental vrma support

yue4u:
add `vrmaSrc` parameter and support vrma loading


packages/vrm-viewer/src/features/vrm-mixin.ts
```diff
@@ -1,6 +1,11 @@
 import { VRM, VRMLoaderPlugin, VRMUtils } from '@pixiv/three-vrm';
+import { createVRMAnimationClip, VRMAnimationLoaderPlugin, VRMLookAtQuaternionProxy } from '@pixiv/three-vrm-animation';
```
yue4u, line 2: I think it's rare to display models as T-pose so it should be ok to always  import these.

packages/vrm-viewer/examples/models/idle_loop.vrma
yue4u, line 1: from https://github.com/pixiv/VRoidHub-API-Example/blob/main/public/idle_loop.vrma

yue4u:
> Error: ../../node_modules/lit/node_modules/@lit/reactive-element/decorators/event-options.d.ts(10,71): error TS2304: Cannot find name 'ClassMethodDecoratorContext'.

CI requires ts 5.0

yue4u:
https://github.com/pixiv/three-vrm/pull/1395#issuecomment-2144469015

PR #1402, opened 2024-05-15
fix: MToon, Fix uv animation transform order

0b5vr:
This fixes the uv animation transform order.

The spec says that the animation should not accelerate over time when we apply uv scroll and rotation over time.

See: https://github.com/vrm-c/vrm-specification/tree/master/specification/VRMC_materials_mtoon-1.0#transform-order

### DIFF
--- a/packages/three-vrm-materials-mtoon/src/shaders/mtoon.frag
+++ b/packages/three-vrm-materials-mtoon/src/shaders/mtoon.frag
@@ -389,10 +389,10 @@ void main() {
       uvAnimMask = texture2D( uvAnimationMaskTexture, uvAnimationMaskTextureUv ).b;
     #endif
 
-    uv = uv + vec2( uvAnimationScrollXOffset, uvAnimationScrollYOffset ) * uvAnimMask;
     float uvRotCos = cos( uvAnimationRotationPhase * uvAnimMask );
     float uvRotSin = sin( uvAnimationRotationPhase * uvAnimMask );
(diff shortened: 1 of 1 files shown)


packages/three-vrm-materials-mtoon/src/shaders/mtoon.frag
```diff
@@ -389,10 +389,10 @@ void main() {
       uvAnimMask = texture2D( uvAnimationMaskTexture, uvAnimationMaskTextureUv ).b;
     #endif
 
-    uv = uv + vec2( uvAnimationScrollXOffset, uvAnimationScrollYOffset ) * uvAnimMask;
     float uvRotCos = cos( uvAnimationRotationPhase * uvAnimMask );
     float uvRotSin = sin( uvAnimationRotationPhase * uvAnimMask );
     uv = mat2( uvRotCos, -uvRotSin, uvRotSin, uvRotCos ) * ( uv - 0.5 ) + 0.5;
+    uv = uv + vec2( uvAnimationScrollXOffset, uvAnimationScrollYOffset ) * uvAnimMask;
```
yue4u, line 395: Reorder to follow the spec.

yue4u (commented):
LGTM. Should this be a breaking change? I think it depends on how many models are affected.

yue4u (approved):

0b5vr:
> Should this be a breaking change? I think it depends on how many models are affected.

I define this as a bugfix instead of a behavior change, since the spec already defines a desired behavior.

PR #1406, opened 2024-05-17
feat: typescript v5

yue4u:
Upgrade to typescript v5 and bump related deps (eslint, jest, typedoc, ...etc)


packages/three-vrm-materials-mtoon/src/utils/getTextureColorSpace.ts
```diff
@@ -1,7 +1,9 @@
 import * as THREE from 'three';
 
 const encodingColorSpaceMap: Record<any, '' | 'srgb'> = {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
```
yue4u, line 4: It's possible to avoid this comment via tweaking eslint config but that needs some regex magic. I think it's easier to disable here.

.eslintrc.json
```diff
@@ -53,6 +53,10 @@
       {
         "selector": "enumMember",
         "format": ["PascalCase"]
+      },
```
yue4u, line 56: for `THREE`, `V1MToonSchema`, `V0VRM` ...

package.json
```diff
@@ -38,11 +38,11 @@
...
-    "ts-jest": "^26.4.4",
-    "ts-node": "^9.0.0",
-    "tslib": "^2.0.3",
-    "typedoc": "^0.25.4",
-    "typescript": "^4.3.5"
+    "ts-jest": "^29.1.2",
+    "ts-node": "^10.9.2",
+    "tslib": "^2.6.2",
+    "typedoc": "^0.25.13",
+    "typescript": "^5.4.5"
```
yue4u, line 45: I think three-vrm still does not require users to use typescript v5 yet after this PR.

yue4u: PR to remove `downlevel-dts` https://github.com/pixiv/three-vrm/pull/1302 . I think that is still valid and can be merged either before or after this PR.

package.json
```diff
@@ -18,18 +18,18 @@
...
-    "@typescript-eslint/eslint-plugin": "^4.6.1",
-    "@typescript-eslint/parser": "^4.6.1",
+    "@rollup/plugin-typescript": "^11.1.6",
+    "@types/jest": "^29.5.12",
+    "@typescript-eslint/eslint-plugin": "^7.9.0",
+    "@typescript-eslint/parser": "^7.9.0",
     "cross-env": "^7.0.2",
     "downlevel-dts": "^0.11.0",
-    "eslint": "^7.12.1",
+    "eslint": "8",
```
0b5vr, line 27: is the version specifier intentional? eslint seems to bump minor frequently so it might be reasonable enough to do this though

yue4u: ESlint v9 is out but not very well supported by plugins for now. Basically this means latest v8 and I don't think there'll be many v8 releases from now on.

0b5vr: sure! if it's intentional it's fine

0b5vr:
merging into `dev-v3`.

PR #1407, opened 2024-05-17
fix: Fill default MToon parameters for v0 VRMs

0b5vr:
Sequel of #1404 

This PR fills default MToon parameters for v0 VRMs.

it's very rare but some models don't include default parameters.


yue4u (approved):
Did not  check the detailed values. (Since there's no documentation those values can only be verified by creating mtoon in unity)

PR #1410, opened 2024-05-20
fix: MToon, Fix normal mapping when other textures are not specified

0b5vr:
See: https://github.com/pixiv/three-vrm/pull/1384#issuecomment-2117311809

Fix normal mapping when other textures are not specified.

This is because `this.normalMap !== null` did not perticipate in the condition of the `MTOON_USE_UV` flag.

Below is the screenshot of `feature-test.html`. Note that the blue ball that only specifies normal map has bumps correctly.

image


ke456-png (approved):
NICE! 👍

PR #1411, opened 2024-05-22
BREAKING: feat: migrate to esbuild

yue4u:
Switch from rollup to esbuild. Enabling esm build by default with cjs fallback

### BREAKING CHANGES

- iife build is no longer provided.


.eslintrc.json
```diff
@@ -9,6 +9,9 @@
     "plugin:@typescript-eslint/recommended",
     "plugin:@typescript-eslint/eslint-recommended"
   ],
+  "globals": {
+    "process": true
+  },
```
yue4u, line 14: We provide this via esbuild define. It makes more sense compare to adding node to env.

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "private": true,
+  "type": "module",
```
yue4u, line 3: Since we are adding `"type": "module"` to packages, `.js` files are treated as esm; therefore using `.mjs` or `.module.js` should not make a difference from the user perspective. I guess `.module.js` is more common in the three ecosystem?

packages/three-vrm-animation/package.json
```diff
@@ -20,6 +20,14 @@
       ]
     }
   },
+  "exports": {
+    ".": {
+      "types": "./types/index.d.ts",
```
yue4u, line 25: [publint](https://github.com/bluwy/publint) says

```
1. pkg.exports["."].types types is interpreted as ESM when resolving with the "require" condition. 
This causes the types to only work when dynamically importing the package, 
even though the package exports CJS. 
Consider splitting out two "types" conditions for "import" and "require", 
and use the .cts extension, e.g. pkg.exports["."].require.types: "./types/index.d.cts"
```

but seperating declarations to `mts` and `cts` does not make many sense to me and causes trouble for building.

packages/three-vrm-animation/package.json
```diff
@@ -31,8 +39,9 @@
     "dev": "cross-env NODE_ENV=development rollup -w -c",
     "clean": "rimraf docs/ lib/ ts*/ types/",
     "build": "yarn build-dev && yarn build-prod && yarn build-types",
-    "build-dev": "cross-env NODE_ENV=development rollup -c",
-    "build-prod": "cross-env NODE_ENV=production rollup -c",
+    "build-js": "node ../../bin/build.mjs",
```
yue4u, line 42: yarn v1 does not support running workspace command. (yarn v2~ / npm /  pnpm have this feature)

packages/three-vrm-animation/package.json
```diff
@@ -20,6 +20,14 @@
       ]
     }
   },
+  "exports": {
+    ".": {
+      "types": "./types/index.d.ts",
+      "import": "./lib/three-vrm-animation.min.mjs",
+      "require": "./lib/three-vrm-animation.min.cjs"
```
yue4u, line 27: not sure if the [`development`](https://webpack.js.org/guides/package-exports/#:~:text=%3A%20%22module%22%2C%0A%20%20%22exports%22%3A%20%7B%0A%20%20%20%20%22-,development,-%22%3A%20%22./index%2Dwith%2Ddevtools.js%22%2C%0A%20%20%20%20%22default) field is really useful. (never used this directly by myself)

.eslintrc.json
0b5vr, line 14: This is for `process.cwd` in `build.mjs`, right?

bin/build.mjs
```diff
@@ -0,0 +1,120 @@
...
+ */
+async function buildPackage(absWorkingDir) {
+  /**
+   * @type {{ name: string, version: string, description: string, module: string }}
+   */
+  const packageJson = JSON.parse(await fs.readFile(path.join(absWorkingDir, './package.json'), 'utf-8'));
+
+  // == constants ====================================================================================
+  /** copyright text */
+  const copyright = '(c) 2020-2024 pixiv Inc.';
```
0b5vr:
```suggestion
  const copyright = '(c) 2019-2024 pixiv Inc.';
```

.eslintrc.json
yue4u: also for `process.env.*`

README.md
```diff
@@ -46,7 +46,7 @@ Code like this:
     "imports": {
       "three": "https://cdn.jsdelivr.net/npm/three@0.164.1/build/three.module.js",
       "three/addons/": "https://cdn.jsdelivr.net/npm/three@0.164.1/examples/jsm/",
-      "@pixiv/three-vrm": "three-vrm.module.js"
+      "@pixiv/three-vrm": "three-vrm.mjs"
```
0b5vr: we also need to change this:

https://github.com/pixiv/three-vrm/pull/1411/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R36

package.json
0b5vr: yeah I lowkey think so > I guess `.module.js` is more common in the three ecosystem?

This is how camera-controls serves its builds, for reference:
https://www.jsdelivr.com/package/npm/camera-controls?tab=files&path=dist

package.json
```diff
@@ -9,6 +10,7 @@
   "scripts": {
     "all": "yarn clean && yarn build && yarn test && yarn lint && yarn docs",
     "clean": "lerna run clean",
+    "dev": "cross-env NODE_ENV=development SERVE=1 node ./bin/build.mjs",
```
0b5vr, line 13: what does it do?

yue4u: `yarn dev packages/three-vrm` to start developing on a package, shorter and faster compared to `NODE_ENV=development SERVE=1 yarn workspace @pixiv/three-vrm build` or `cd`ing

package.json
yue4u: changed to `module.js` in [2782638](https://github.com/pixiv/three-vrm/pull/1411/commits/27826382b8df2effd8ad34d9e289d2ac0c796113)

0b5vr: much fewer diff!

package.json
0b5vr: who specifies the `packages/three-vrm` in this command?

yue4u: `packages/three-vrm` is the arg (`process.argv[2]`) passed to cli. See https://github.com/pixiv/three-vrm/pull/1411/files#diff-a9c6b99e70a85f0bb0fd35f083215fec11ae9bddf6e29853d04831ce1b813ec9R14

yue4u: running `yarn dev` without arg will error out but I don't think that case worth handling

0b5vr: ah got it got it, you said `yarn dev packages/three-vrm`. looks good

0b5vr (approved):
Looks good, thanks!

PR #1412, opened 2024-05-24
`dev-v3`

0b5vr:
This is the development branch of the next major version.


0b5vr:
We need to check if the WebGPU one works properly on all major devices before we ship this.

PR #1413, opened 2024-05-24
Fix leftover rollup stuff

0b5vr:
- Remove leftover rollup deps.
- Change `dev` commands of each package from the rollup invocation to the esbuild script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
   },
   "devDependencies": {
     "@gltf-transform/core": "^3.7.4",
-    "@rollup/plugin-typescript": "^11.1.6",
     "@types/jest": "^29.5.12",
     "@types/node": "^20.12.12",
     "@typescript-eslint/eslint-plugin": "^7.9.0",
--- a/packages/three-vrm-animation/package.json
+++ b/packages/three-vrm-animation/package.json
@@ -36,7 +36,7 @@
   "scripts": {
     "version": "yarn all",
     "all": "yarn lint && yarn clean && yarn build && yarn docs",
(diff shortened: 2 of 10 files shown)


packages/three-vrm-animation/package.json
```diff
@@ -36,7 +36,7 @@
   "scripts": {
     "version": "yarn all",
     "all": "yarn lint && yarn clean && yarn build && yarn docs",
-    "dev": "cross-env NODE_ENV=development rollup -w -c",
+    "dev": "cross-env NODE_ENV=development SERVE=1 node ../../bin/build.mjs",
```
yue4u: dev command can replaced with `SERVE=1 yarn build-dev`?

If this is stil necessary I'd perfer 
`cross-env SERVE=1 yarn build-dev` like `build-dev` to reuse command definition since `yarn build-js` can be move to workspace root once package manager support.

package.json
```diff
@@ -20,7 +20,6 @@
   },
   "devDependencies": {
     "@gltf-transform/core": "^3.7.4",
-    "@rollup/plugin-typescript": "^11.1.6",
```
yue4u, line 23: good catch

packages/three-vrm-animation/package.json
0b5vr: true!

0b5vr: addressed @ https://github.com/pixiv/three-vrm/pull/1413/commits/46b2b2d53b68e380a0fabd2ad386aae283fffa7f

yue4u (approved):

PR #1415, opened 2024-05-28
BREAKING: fix: default needThumbnailImage to false

yue4u:
make needThumbnailImage default to false

### DIFF
--- a/packages/three-vrm-core/src/meta/VRMMetaLoaderPlugin.ts
+++ b/packages/three-vrm-core/src/meta/VRMMetaLoaderPlugin.ts
@@ -22,7 +22,7 @@ export class VRMMetaLoaderPlugin implements GLTFLoaderPlugin {
 
   /**
    * If `false`, it won't load its thumbnail image ({@link VRM1Meta.thumbnailImage}).
-   * `true` by default.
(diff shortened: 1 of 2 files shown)


packages/three-vrm-core/src/meta/VRMMetaLoaderPlugin.ts
```diff
@@ -48,7 +48,7 @@ export class VRMMetaLoaderPlugin implements GLTFLoaderPlugin {
   public constructor(parser: GLTFParser, options?: VRMMetaLoaderPluginOptions) {
     this.parser = parser;
 
-    this.needThumbnailImage = options?.needThumbnailImage ?? true;
+    this.needThumbnailImage = options?.needThumbnailImage ?? false;
```
yue4u, line 51: I've been wanted to fix this for a while. As a breaking and it's only doable during major update.

0b5vr:
heyyy make sure you also change this line

https://github.com/pixiv/three-vrm/pull/1415/files#diff-e72149acd64a2275382fc08b7601eabc422e3dac26ad45b80853b1cff292b4bbR25

0b5vr (approved):

PR #1420, opened 2024-06-12
docs: Add doc comments and README descriptions about WebGPU MToon stuff

0b5vr:
Changes of imports in this PR are all type imports and only used for doc comments

I have to add support for v0CompatShade to the WebGPU one later

I think it's high time to remove the v1 notice on the root README

I thought that we could specify more simpler option like `nodeMaterial: true` to the option of `VRMLoaderPlugin` while writing these docs

### Points need review

- [ ] Do you think it's easy to follow when you try to use three-vrm in WebGPURenderer?

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,9 @@ loader.load(
 Starting from v3, we provide [WebGPURenderer](https://github.com/mrdoob/three.js/blob/master/examples/jsm/renderers/webgpu/WebGPURenderer.js) compatibility.
 To use three-vrm with WebGPURenderer, specify the WebGPU-compatible `MToonNodeMaterialLoaderPlugin` for the `mtoonMaterialPlugin` option of `VRMLoaderPlugin`.
 
+`MToonNodeMaterial` only supports Three.js r161 or later.
+The NodeMaterial system of Three.js is still under development, so we may break compatibility with older versions of Three.js more frequently than other parts of three-vrm.
+
 ```js
 import { VRMLoaderPlugin, MToonNodeMaterialLoaderPlugin } from '@pixiv/three-vrm';
 
--- a/README.md
+++ b/README.md
@@ -14,16 +14,6 @@ Use [VRM](https://vrm.dev/) on [three.js](https://threejs.org/)
 
 [API Reference](https://pixiv.github.io/three-vrm/packages/three-vrm/docs)
 
-## v1
-
-**three-vrm v1 has been released!**
-
-three-vrm v1 supports [VRM1.0](https://vrm.dev/vrm1/), which is a new version of VRM format (the previous version of VRM is also supported, don't worry!).
-It also adopts the GLTFLoader plugin system which is a relatively new feature of GLTFLoader.
-
-There are a lot of breaking changes!
-See [the migration guide](https://github.com/pixiv/three-vrm/blob/dev/docs/migration-guide-1.0.md) for more info.
-
 ## How to Use
 
 ### from HTML
@@ -136,6 +126,32 @@ loader.load(
   // called when loading has errors
   (error) => console.error(error),
 );
+```
+
+### Use with WebGPURenderer
+
+Starting from v3, we provide a WebGPURenderer compatibility.
+For now, in order to use with WebGPURenderer, you need to specify a WebGPU compatible MToon loader plugin to the option `mtoonMaterialPlugin` of `VRMLoaderPlugin`.
+
+```js
+import { VRMLoaderPlugin, MToonNodeMaterialLoaderPlugin } from '@pixiv/three-vrm';
+
+// ...
+
+// Register a VRMLoaderPlugin
+loader.register((parser) => {
+
+  // create a WebGPU compatible MToon loader plugin
+  const mtoonMaterialPlugin = new MToonNodeMaterialLoaderPlugin(parser);
+
+  return new VRMLoaderPlugin(parser, {
+
+    // Specify the MToon loader plugin to use in the VRMLoaderPlugin instance
+    mtoonMaterialPlugin,
+
+  });
+
+});
 ```
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ loader.load(
 
 ### Use with WebGPURenderer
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -138,6 +128,32 @@ loader.load(
...
+// Register a VRMLoaderPlugin
+loader.register((parser) => {
+
+  // create a WebGPU compatible MToon loader plugin
+  const mtoonMaterialPlugin = new MToonNodeMaterialLoaderPlugin(parser);
+
+  return new VRMLoaderPlugin(parser, {
+
+    // Specify the MToon loader plugin to use in the VRMLoaderPlugin instance
+    mtoonMaterialPlugin,
```
yue4u, line 153: Question: Can this be detected inside three-vrm? Should we mention `renderer.isWebGPURenderer`?

packages/three-vrm-materials-mtoon/src/MToonMaterialLoaderPlugin.ts
```diff
@@ -1,19 +1,26 @@
...
-import type { GLTF, GLTFLoaderPlugin, GLTFParser } from 'three/examples/jsm/loaders/GLTFLoader.js';
+import type { GLTF, GLTFLoader, GLTFLoaderPlugin, GLTFParser } from 'three/examples/jsm/loaders/GLTFLoader.js';
 import type { MToonMaterialParameters } from './MToonMaterialParameters';
 import { MToonMaterialOutlineWidthMode } from './MToonMaterialOutlineWidthMode';
 import { GLTFMToonMaterialParamsAssignHelper } from './GLTFMToonMaterialParamsAssignHelper';
 import { MToonMaterialLoaderPluginOptions } from './MToonMaterialLoaderPluginOptions';
 import type { MToonMaterialDebugMode } from './MToonMaterialDebugMode';
 import { GLTF as GLTFSchema } from '@gltf-transform/core';
 import { MToonMaterial } from './MToonMaterial';
+import { MToonNodeMaterialLoaderPlugin } from './nodes/MToonNodeMaterialLoaderPlugin';
```
yue4u:
```suggestion
import type { MToonNodeMaterialLoaderPlugin } from './nodes/MToonNodeMaterialLoaderPlugin';
```

packages/three-vrm-materials-mtoon/src/nodes/MToonNodeMaterial.ts
```diff
@@ -43,6 +44,14 @@ import { mtoonParametricRim } from './mtoonParametricRim';
...
 // import { float } from '../shadernode/ShaderNode.js';
 
+/**
+ * MToon is a material specification that has various features.
+ * The spec and implementation are originally founded for Unity engine and this is a port of the material.
+ *
+ * This material is a NodeMaterial variant of {@link MToonMaterial}.
+ *
+ * See: https://github.com/Santarh/MToon
+ */
```
yue4u, line 54: 👍

README.md
0b5vr: I think `VRMLoaderPlugin` doesn't have an access to `renderer`.

yue4u: > Can this be detected inside three-vrm
So this is no.

> Should we mention

I mean if the renderer may be created before `VRMLoaderPlugin` registration and exist somewhere in current execution context, user can figure out what to use via `isWebGPURenderer`.

0b5vr: Renderer and VRMLoaderPlugin should be loosely coupled I think. What does the code you expected look like?

yue4u: Something like
```ts
const mtoonMaterialPlugin = renderer.isWebGPURenderer === true 
  ? new MToonNodeMaterialLoaderPlugin(parser)
  : undefined
```

or just a comment

```
// assert renderer.isWebGPURenderer is true
```

I know this is totally optional but may provide some useful info to users.

README.md
```diff
@@ -138,6 +128,32 @@ loader.load(
...
+```js
+import { VRMLoaderPlugin, MToonNodeMaterialLoaderPlugin } from '@pixiv/three-vrm';
+
+// ...
+
+// Register a VRMLoaderPlugin
+loader.register((parser) => {
+
+  // create a WebGPU compatible MToon loader plugin
+  const mtoonMaterialPlugin = new MToonNodeMaterialLoaderPlugin(parser);
```
0b5vr, line 148:
```suggestion
  const mtoonMaterialPlugin = renderer.isWebGPURenderer === true
    ? new MToonNodeMaterialLoaderPlugin(parser)
    : undefined; // fallbacks to the default `MToonMaterialLoaderPlugin`
```

README.md
0b5vr: I see I see.

https://github.com/pixiv/three-vrm/pull/1420/files#r1642356443 👈 Does this look good?

README.md
0b5vr: This is kinda redundant considering that the document already says that it's for `WebGPURenderer`.

packages/three-vrm-materials-mtoon/README.md
```diff
@@ -10,9 +10,11 @@ MToon (toon material) module for @pixiv/three-vrm
 
 ## WebGPU Support
 
-This module supports [WebGPURenderer](https://github.com/mrdoob/three.js/blob/master/examples/jsm/renderers/webgpu/WebGPURenderer.js).
+Starting from v3, we provide [WebGPURenderer](https://github.com/mrdoob/three.js/blob/master/examples/jsm/renderers/webgpu/WebGPURenderer.js) compatibility.
+To use three-vrm with WebGPURenderer, specify the WebGPU-compatible `MToonNodeMaterialLoaderPlugin` for the `mtoonMaterialPlugin` option of `VRMLoaderPlugin`.
 
-To use MToon with `VRMLoaderPlugin` in WebGPURenderer, specify the WebGPU-compatible `MToonNodeMaterialLoaderPlugin` for the `mtoonMaterialPlugin` option of `VRMLoaderPlugin`.
+`MToonNodeMaterial` only supports Three.js r161 or later.
+The NodeMaterial system of Three.js is still under development, so we may break compatibility with older versions of Three.js more frequently than other parts of three-vrm.
```
ke456-png, line 17: Thanks!

PR #1424, opened 2024-06-20
feat: MToonNodeMaterial, move opaque stuff from setupOutput to setupDiffuseColor

0b5vr:
following the Three.js side change

See: https://github.com/mrdoob/three.js/pull/28646


ke456-png (approved):
thanks!

PR #1426, opened 2024-06-27
fix: re-enable thumbnail loading in examples

yue4u:
ref: https://github.com/pixiv/three-vrm/pull/1415
re-enable thumbnail loading in examples


0b5vr:
Thanks!

PR #1427, opened 2024-06-27
fix: seperate nodes entry for backward compatibility

yue4u:
This PR attempts to address compatibility issues with three before r161.

Removed `export * from './nodes'` in `three-vrm-materials-mtoon` and created seperated `./nodes` entry in both `three-vrm-materials-mtoon` and `three-vrm`


tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "es6",
     "module": "es6",
     "lib": ["es2018", "dom"],
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
```
yue4u, line 6: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#--moduleresolution-bundler
Expect this not to affect end user

packages/three-vrm/src/nodes/index.ts
```diff
@@ -0,0 +1 @@
+export * from '@pixiv/three-vrm-materials-mtoon/nodes';
```
yue4u, line 1: Should we add some three version checks and log info here or inside `@pixiv/three-vrm-materials-mtoon/nodes`? IMO it's not necessary

README.md
```diff
@@ -92,7 +92,7 @@ Code like this:
 
 ```javascript
-import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader';
+import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js';
```
yue4u, line 95: align to three export

bin/build.mjs
```diff
@@ -55,16 +55,46 @@ async function buildPackage(absWorkingDir) {
...
+    /**
+     * @param {string} base
+     */
+    const addSuffix = (base) => {
+      let outFilename = base + (format === 'esm' ? '.module' : '');
+      return outFilename + (DEV ? '' : '.min');
+    };
+    let outFilename = addSuffix(filename);
+
+    const extraEntryPoints = ['@pixiv/three-vrm', '@pixiv/three-vrm-materials-mtoon'].includes(packageJson.name)
```
yue4u: doing this here to modify cjs,esm outputs altogether

bin/build.mjs
```diff
@@ -55,16 +55,46 @@ async function buildPackage(absWorkingDir) {
...
     return {
       entryPoints: {
         [outFilename]: 'src/index.ts',
+        ...extraEntryPoints,
       },
       format,
     };
   };
 
+  /**
```
yue4u: Coundn't find a better way to do this. IMO it's too wasteful to bundle shaders and everything from three-vrm-materials-mtoon main entry again

tsconfig.json
0b5vr: why is the first letter capitalized though?

packages/three-vrm/src/nodes/index.ts
0b5vr: we already have a warning inside `MToonNodeMaterial` so I think it's not needed

tsconfig.json
yue4u: From autocompletion. (guess this is the default)
Screenshot 2024-06-28 at 13 32 36

packages/three-vrm/src/nodes/index.ts
yue4u: I think MToonNodeMaterial'constructor will never be executed if `three/addons/nodes/Nodes` not exist?

bin/build.mjs
0b5vr: This section looks surprisingly cryptic compared to what it actually does 😩

bin/build.mjs
```diff
@@ -55,16 +55,46 @@ async function buildPackage(absWorkingDir) {
...
+     */
+    const addSuffix = (base) => {
+      let outFilename = base + (format === 'esm' ? '.module' : '');
+      return outFilename + (DEV ? '' : '.min');
+    };
+    let outFilename = addSuffix(filename);
+
+    const extraEntryPoints = ['@pixiv/three-vrm', '@pixiv/three-vrm-materials-mtoon'].includes(packageJson.name)
+      ? { [addSuffix('nodes/index')]: 'src/nodes/index.ts' }
+      : {};
```
0b5vr:
```suggestion
    const needsNodesBuild = ['@pixiv/three-vrm', '@pixiv/three-vrm-materials-mtoon'].includes(packageJson.name);
    const extraEntryPoints = needsNodesBuild
      ? { [addSuffix('nodes/index')]: 'src/nodes/index.ts' }
      : {};
```

bin/build.mjs
```diff
@@ -55,16 +55,46 @@ async function buildPackage(absWorkingDir) {
    * @param {'esm' | 'cjs'} format
    */
   const entryPoints = (format) => {
-    let outFilename = filename + (format === 'esm' ? '.module' : '');
-    outFilename += DEV ? '' : '.min';
+    /**
+     * @param {string} base
+     */
```
0b5vr:
```suggestion
    /**
     * Add suffix according to the format to the given basename of the output file.
     * @param {string} base Basename of the output file
     * @returns {string} `<base>[.module][.min]`
     */
```

bin/build.mjs
```diff
@@ -55,16 +55,46 @@ async function buildPackage(absWorkingDir) {
...
+    };
+    let outFilename = addSuffix(filename);
+
+    const extraEntryPoints = ['@pixiv/three-vrm', '@pixiv/three-vrm-materials-mtoon'].includes(packageJson.name)
+      ? { [addSuffix('nodes/index')]: 'src/nodes/index.ts' }
+      : {};
+
     return {
       entryPoints: {
         [outFilename]: 'src/index.ts',
```
0b5vr: I think we can directly put `addSuffix(filename)` here as `extraEntryPoints` does

bin/build.mjs
0b5vr: I will find a way to quit `MToonNodeMaterialLoaderPlugin`. Maybe having a factory pattern or something would help this

tsconfig.json
0b5vr: huh. It's kinda weird for me but it's ok.

bin/build.mjs
yue4u: removed in 6d40b26a5d52d82c00d40b66f72874e5c507744b

bin/build.mjs
yue4u: [e2e96c9](https://github.com/pixiv/three-vrm/pull/1427/commits/e2e96c9a27992cf7001ba81e27350643ae97dc7f)

yue4u:
webgpu-dnd example now produces a lot of errors when using with an order version of three. Can we do something about this?

Screenshot 2024-07-01 at 15 35 05

0b5vr:
> webgpu-dnd example now produces a lot of errors when using with an order version of three. Can we do something about this?

`WebGPURenderer.renderAsync` has been introduced in r161. I believe it's fine in all versions of r161 or later

See: https://github.com/mrdoob/three.js/pull/27597

packages/three-vrm/src/nodes/index.ts
0b5vr: Ahh, yes it's possible to cause an error before `MToonNodeMaterial` instantiates.

Does this help?

image

0b5vr (approved):

PR #1428, opened 2024-06-27
Specify more appropriate Three.js version in peerDependencies 

0b5vr:
- Specify more appropriate Three.js version in peerDependencies.
    - npm users suffer from unmatched peerDependencies (the `--legacy-peer-deps` option would be required). This should be fixed by this PR.
- Removed old compat codes that were for pre-r137


0b5vr:
after #1427 we are actually able to make it work on r137 and later

yue4u (approved):

PR #1431, opened 2024-07-01
fix: Add new shadowIntensity value to mtoon shader

ashconnell:
Three 166 introduces a new light.shadowIntensity value (https://github.com/mrdoob/three.js/pull/28588) which breaks three-vrm as the shader won't compile.

This PR updates the MToon fragment shader to use the new value.


0b5vr:
@ashconnell Thank you for the contribution!

We usually provide broader compatibility against past Three.js versions, so compat codes are mandatory in MToon shader codes.
Would you add a compat code following the code example below?
You would probably use the expression `THREE_VRM_THREE_VERSION >= 166`.

https://github.com/pixiv/three-vrm/blob/9e16c3855a5b441cca0f11480c4003de25d2943c/packages/three-vrm-materials-mtoon/src/shaders/mtoon.frag#L653-L660

After changing the code, please make sure that the code is working correctly on both r165 and r166.

0b5vr:
@ashconnell Hello, could you make the change I mentioned above?

0b5vr:
Superseded by #1441

PR #1232, opened 2023-06-07
Bump Three.js to r153

0b5vr:
### Description

- 📦 Bumped Three.js to r153.
    - See: https://github.com/mrdoob/three.js/releases/tag/r153
- ⬅ Add compat code for pre-152, related to color space.
    - `texture.encoding` has been renamed to `texture.colorSpace` in r152.
- 🚨⬅ BREAKING: remove pre-r137 compat codes.
    - This drops the support of pre-r137 (Jan 29, 2022).
- 📄 Remove color space specification from example codes.
- 📄 Update the 1.0 migration guide to follow the r152 change.

This PR will bump the MAJOR version.

PR #1445, opened 2024-07-22
fix: restore alpha test workaround

yue4u:
Revert changes in https://github.com/pixiv/three-vrm/pull/1428/files#diff-e2dae1fd189076ccced184108018ba54632410b52ef00fe0f6bbc43673ef6733L590-L596

Since we only supports r137+ so the version check is redundant.


ke456-png:
[Memo] 
three-vrm v2.0 or later should support Three.js r137 or later

> 🚨 BREAKING CHANGES
> https://github.com/pixiv/three-vrm/pull/1232: This drops support of Three.js r136 and prior.
(https://github.com/pixiv/three-vrm/releases/tag/v2.0.0)

ke456-png (approved):
NICE! 👍

PR #1449, opened 2024-07-25
feat: VRMAnimation, emit a warning if the rest hips potision is approximately zero

0b5vr:
VRMAnimationLoaderPlugin now emits a warning if the rest hips potision is approximately zero.

The animation violates the VRM T-pose the hips potision is approximately zero.

Based on the spec: https://github.com/vrm-c/vrm-specification/tree/master/specification/VRMC_vrm_animation-1.0#humanoid


0b5vr:
just an idea: We might want to create a validator that checks VRM T-pose, that can be used for both VRM and VRMA

ke456-png (approved):
NICE!

Issue #1447, opened 2024-07-24
Model moving down when sitting mixamo animation applied


DENNKA:
https://pixiv.github.io/three-vrm/packages/three-vrm/examples/humanoidAnimation/index.html
Reproduce:
- Sitting animation (all is ok)
- Load any another (e. g. same)  
  
Model's legs below ground. 
I'm newbie i don't know how to fix it.
![Screenshot](https://github.com/user-attachments/assets/3a0dce3f-b686-4a2a-b475-518f9ca896d7)
[Sitting And Rubbing Arm.zip](https://github.com/user-attachments/files/16367711/Sitting.And.Rubbing.Arm.zip)

0b5vr:
Reproduced. This is the bug of the example.

PR #1455, opened 2024-08-02
Fix humanoidAnimation example

ke456-png:
Fix issue #1447.

When additional motion was loaded, the root motion was out of alignment.


0b5vr (approved):
Working excellent

PR #1462, opened 2024-08-09
fix: change VRMSpringBoneManager circular dependency error to warning

yue4u:
Change VRMSpringBoneManager circular dependency error to warning.
After this PR three-vrm will try the best to render models contain this issue instead of error out.


packages/three-vrm-springbone/src/VRMSpringBoneManager.ts
```diff
@@ -7,6 +7,8 @@ import { traverseChildrenUntilConditionMet } from './utils/traverseChildrenUntil
 
 export class VRMSpringBoneManager {
   private _joints = new Set<VRMSpringBoneJoint>();
+  private _hasWarnedCircularDependency = false;
```
0b5vr, line 10: warn once per VRMSpringBoneManager (= a model). Good

0b5vr (approved):
Thank you!

PR #1465, opened 2024-08-19
Bump Three.js to r168

0b5vr:
### Description

- Bump Three.js to r168
- Add a compat to MToonNodeMaterial `setupNormal`
- Add a compat to `tslFn` calling
- `normalMap` is no longer a node element, we use this as an individual node instead

See each commits for details.

### Points need review

- [x] Does it work on r167?
- [x] Does it work on r168?


0b5vr:
We are going to wait `@types/three`

0b5vr:
Ready to merge, Working fine in (r167, r168) x (WebGL, WebGPU) matrix

Issue #1469, opened 2024-08-21
RenderOrder in ‘VRM0 UnlitTransparentZWrite’ is wrong.


ke456-png:
The following areas need to be corrected.

https://github.com/pixiv/three-vrm/blob/cedeb03f26ebb18e73e54e5c781e3ead9e55d93c/packages/three-vrm-materials-v0compat/src/VRMMaterialsV0CompatPlugin.ts#L287-L290

https://github.com/pixiv/three-vrm/blob/cedeb03f26ebb18e73e54e5c781e3ead9e55d93c/packages/three-vrm-materials-v0compat/src/VRMMaterialsV0CompatPlugin.ts#L359-L361

0b5vr:
resolved in #1473

PR #1473, opened 2024-08-23
fix UnlitTransparentZWrite

ke456-png:
#1469  Fixed the RenderQue of UnlitTransparentZWrite.


0b5vr (approved):
Thank you. Tested on my end and working fine. The logic here is kinda catastrophic but we don't touch here often

PR #1485, opened 2024-08-29
Consider `isBinary` in weight for override amounts

### LINKED ISSUE #1484: Incorrect override amounts when using binary expressions
When an expression is set to binary with `isBinary: true`, and these expressions also have blended overrides (e.g. `overrideLookAt: "blend"`), their weight is blended continuously instead of discretely when the corresponding expression group (e.g. gaze expression) has nonzero weight. I expect binary expressions to maintain the sharp 0.5 threshold whether using "blend" overrides or not.

I believe this stems from the `overrideBlinkAmount` / `overrideLookAtAmount` / `overrideMouthAmount` calculations not taking into account the `isBinary` flag, e.g.:

https://github.com/pixiv/three-vrm/blob/7a35f7ee456136602d7bc12450b6fe9c247d7f5d/packages/three-vrm-core/src/expressions/VRMExpression.ts#L61-L69

Changing the expression in the 'blend' case to `this.isBinary ? (this.weight <= 0.5 ? 0.0 : 1.0) : this.weight` as is [done elsewhere in the applyWeight logic](https://github.com/pixiv/three-vrm/blob/7a35f7ee456136602d7bc12450b6fe9c247d7f5d/packages/three-vrm-core/src/expressions/VRMExpression.ts#L115) produces the expected discrete blending behavior.

I wonder if this also affects the 'block' override mode, as the binary expression will begin blocking as soon as it has nonzero weight, even though it doesn't visually apply that weight until after the 0.5 weight threshold. So perhaps the 'block' case should look more like:
```ts
const actualWeight = this.isBinary ? (this.weight <= 0.5 ? 0.0 : 1.0) : this.weight;
return 0.0 < actualWeight ? 1.0 : 0.0;
```

mattrossman:
Resolves #1484

### DIFF
--- a/packages/three-vrm-core/src/expressions/VRMExpression.ts
+++ b/packages/three-vrm-core/src/expressions/VRMExpression.ts
@@ -46,9 +46,10 @@ export class VRMExpression extends THREE.Object3D {
    */
   public get overrideBlinkAmount(): number {
     if (this.overrideBlink === 'block') {
-      return 0.0 < this.weight ? 1.0 : 0.0;
+      const actualWeight = this.isBinary ? (this.weight <= 0.5 ? 0.0 : 1.0) : this.weight;
+      return 0.0 < actualWeight ? 1.0 : 0.0;
     } else if (this.overrideBlink === 'blend') {
-      return this.weight;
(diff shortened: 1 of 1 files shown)


0b5vr:
Superseded by #1489

PR #1489, opened 2024-09-06
change: modify behavior of expressions, interaction between isBinary and override

0b5vr:
This PR updates the behavior of expressions.

- When an expression with isBinary overrides other expressions, it uses the binary output value to affect other expressions
- When an expression with isBinary is overridden by other expressions, the expression is completely suppressed if the effect received is greater than 0.0
- This PR also adds VRMExpression tests to confirm the desired behavior.

### Related link

- The original discussion: https://github.com/pixiv/three-vrm/issues/1484
- The spec change: vrm-c/vrm-specification#487

### DIFF
--- a/packages/three-vrm-core/src/expressions/VRMExpression.ts
+++ b/packages/three-vrm-core/src/expressions/VRMExpression.ts
@@ -92,7 +92,7 @@ export class VRMExpression extends THREE.Object3D {
    */
   public get outputWeight(): number {
     if (this.isBinary) {
-      return this.weight >= 0.5 ? 1.0 : 0.0;
+      return this.weight > 0.5 ? 1.0 : 0.0;
     }
 
     return this.weight;
--- a/packages/three-vrm-core/src/expressions/VRMExpression.ts
+++ b/packages/three-vrm-core/src/expressions/VRMExpression.ts
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import { VRMExpressionBind } from './VRMExpressionBind';
 import type { VRMExpressionOverrideType } from './VRMExpressionOverrideType';
+import type { VRMExpressionManager } from './VRMExpressionManager';
 
 // animationMixer の監視対象は、Scene の中に入っている必要がある。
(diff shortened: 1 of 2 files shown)


packages/three-vrm-core/src/expressions/VRMExpression.ts
```diff
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import { VRMExpressionBind } from './VRMExpressionBind';
 import type { VRMExpressionOverrideType } from './VRMExpressionOverrideType';
+import type { VRMExpressionManager } from './VRMExpressionManager';
```
0b5vr, line 4: `VRMExpressionManager` is only referred from a doc comment

packages/three-vrm-core/src/expressions/VRMExpression.ts
```diff
@@ -74,14 +79,25 @@ export class VRMExpression extends THREE.Object3D {
...
       return 0.0;
     }
   }
 
+  /**
+   * An output weight of this expression, considering the {@link isBinary}.
+   */
+  public get outputWeight(): number {
+    if (this.isBinary) {
+      return this.weight >= 0.5 ? 1.0 : 0.0;
```
ke456-png:
```suggestion
      return this.weight > 0.5 ? 1.0 : 0.0;
```

> isBinary : A value greater than 0.5 is 1.0, otherwise 0.0
> https://github.com/vrm-c/vrm-specification/blob/master/specification/VRMC_vrm-1.0/expressions.md#expression-specification

ke456-png: Please also modify the test cases.

ke456-png (changes requested):

packages/three-vrm-core/src/expressions/VRMExpression.ts
0b5vr: true! I misunderstood the spec.

ke456-png (approved):
NICE!

PR #1492, opened 2024-09-13
fix: Fix screenCoordinates outline rendering in WebGL

0b5vr:
### Description

- ea1e946: This commit adds new materials with outline to `feature-test.html` and `webgpu-feature-test.html`.
    - I suspect that we will have an issue that outline does not appear in r169 (next) WebGPURenderer. This commitment is to prepare to fix the issue.
    - I planned to just do this change in this PR but I spotted an issue I described below,,,
- cd6c501: This commit fixes the `screenCoordinates` outline in WebGL.

The screenshot of the new `feature-test.html`:

image

### DIFF
--- a/packages/three-vrm-materials-mtoon/src/shaders/mtoon.vert
+++ b/packages/three-vrm-materials-mtoon/src/shaders/mtoon.vert
@@ -92,26 +92,21 @@ void main() {
 
   vViewPosition = - mvPosition.xyz;
 
-  float outlineTex = 1.0;
-
   #ifdef OUTLINE
+    float worldNormalLength = length( transformedNormal );
+    vec3 outlineOffset = outlineWidthFactor * worldNormalLength * objectNormal;
+
     #ifdef USE_OUTLINEWIDTHMULTIPLYTEXTURE
       vec2 outlineWidthMultiplyTextureUv = ( outlineWidthMultiplyTextureUvTransform * vec3( vUv, 1 ) ).xy;
-      outlineTex = texture2D( outlineWidthMultiplyTexture, outlineWidthMultiplyTextureUv ).g;
-    #endif
-
-    #ifdef OUTLINE_WIDTH_WORLD
-      float worldNormalLength = length( transformedNormal );
-      vec3 outlineOffset = outlineWidthFactor * outlineTex * worldNormalLength * objectNormal;
-      gl_Position = projectionMatrix * modelViewMatrix * vec4( outlineOffset + transformed, 1.0 );
+      float outlineTex = texture2D( outlineWidthMultiplyTexture, outlineWidthMultiplyTextureUv ).g;
+      outlineOffset *= outlineTex;
     #endif
 
     #ifdef OUTLINE_WIDTH_SCREEN
-      vec3 clipNormal = ( projectionMatrix * modelViewMatrix * vec4( objectNormal, 0.0 ) ).xyz;
-      vec2 projectedNormal = normalize( clipNormal.xy );
-      projectedNormal.x *= projectionMatrix[ 0 ].x / projectionMatrix[ 1 ].y;
-      gl_Position.xy += 2.0 * outlineWidthFactor * outlineTex * projectedNormal.xy;
+      outlineOffset *= vViewPosition.z / projectionMatrix[ 1 ].y;
     #endif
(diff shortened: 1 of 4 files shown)


0b5vr:
> I suspect that we will have an issue that outline does not appear in r169 (next) WebGPURenderer. This commitment is to prepare to fix the issue.

This was because this line fails if we load both `three.module.js` and `three.webgpu.js` in a catastrophic way.

https://github.com/pixiv/three-vrm/blob/cd6c501ea829816e46f9b5cf00f26a7487eaef40/packages/three-vrm-materials-mtoon/src/MToonMaterialLoaderPlugin.ts#L315

And this easily happens when we write our importmap in this way:

```html
<script type="importmap">
  {
    "imports": {
      "three": "https://cdn.jsdelivr.net/npm/three@0.168.0/build/three.module.js",
      "three/webgpu": "https://cdn.jsdelivr.net/npm/three@0.168.0/build/three.webgpu.js",
      // ...
    }
  }
</script>
```

maybe we should consider changing the `instanceof` line to more unsophisticated way? I think the Three.js codebase actually doesn't use `instanceof` against any Three.js classes.

packages/three-vrm-materials-mtoon/src/shaders/mtoon.vert
```diff
@@ -92,27 +92,22 @@ void main() {
...
   #ifdef OUTLINE
+    float worldNormalLength = length( transformedNormal );
+    vec3 outlineOffset = outlineWidthFactor * worldNormalLength * objectNormal;
+
     #ifdef USE_OUTLINEWIDTHMULTIPLYTEXTURE
       vec2 outlineWidthMultiplyTextureUv = ( outlineWidthMultiplyTextureUvTransform * vec3( vUv, 1 ) ).xy;
-      outlineTex = texture2D( outlineWidthMultiplyTexture, outlineWidthMultiplyTextureUv ).g;
-    #endif
-
-    #ifdef OUTLINE_WIDTH_WORLD
```
yue4u, line 103: so [OUTLINE_WIDTH_WORLD](https://github.com/pixiv/three-vrm/blob/c30792c30865de5af7408fd8d0145cc7bc06c716/packages/three-vrm-materials-mtoon/src/MToonMaterial.ts#L609) is no longer needed?

packages/three-vrm-materials-mtoon/src/shaders/mtoon.vert
```diff
@@ -92,27 +92,22 @@ void main() {
...
+      float outlineTex = texture2D( outlineWidthMultiplyTexture, outlineWidthMultiplyTextureUv ).g;
+      outlineOffset *= outlineTex;
     #endif
 
     #ifdef OUTLINE_WIDTH_SCREEN
-      vec3 clipNormal = ( projectionMatrix * modelViewMatrix * vec4( objectNormal, 0.0 ) ).xyz;
-      vec2 projectedNormal = normalize( clipNormal.xy );
-      projectedNormal.x *= projectionMatrix[ 0 ].x / projectionMatrix[ 1 ].y;
-      gl_Position.xy += 2.0 * outlineWidthFactor * outlineTex * projectedNormal.xy;
+      outlineOffset *= vViewPosition.z / projectionMatrix[ 1 ].y;
```
yue4u, line 106: outlineOffset has been fixed and using the same logic in tsl

packages/three-vrm-materials-mtoon/src/shaders/mtoon.vert
0b5vr: Yes, while I agree that it is too implicit to acknowledge that `OUTLINE && !OUTLINE_WIDTH_SCREEN == OUTLINE_WIDTH_WORLD`.

0b5vr: Ended up removing the line @ https://github.com/pixiv/three-vrm/pull/1492/commits/1426a6107b5dfa604203537f0aa6346539ecc96d

yue4u: Maybe leave a comment or something in shader code would help future us? Anyway LGTM.

yue4u (approved):

PR #1515, opened 2024-10-28
fix: Fallback to 'auto' instead of 'both' for MeshAnnotation

mrxz:
The [specification](https://github.com/vrm-c/vrm-specification/blob/master/specification/VRMC_vrm-1.0/firstPerson.md#when-meshannotation-does-not-exist) states that the fallback value for mesh annotations should be `auto`. In case there is no `firstPerson` property, it should still be treated as `auto`. I noticed three-vrm instead used `both` as fallback. This PR changes the fallback value from `both` to `auto` and ensures this is done whether or not there is a `firstPerson` property.

I could not find out if this applies to VRM0 as well, so I've left `_v0Import` unchanged.

### DIFF
--- a/packages/three-vrm-core/src/firstPerson/VRMFirstPersonLoaderPlugin.ts
+++ b/packages/three-vrm-core/src/firstPerson/VRMFirstPersonLoaderPlugin.ts
@@ -90,20 +90,15 @@ export class VRMFirstPersonLoaderPlugin implements GLTFLoaderPlugin {
     }
 
     const schemaFirstPerson = extension.firstPerson;
-    if (!schemaFirstPerson) {
-      return null;
-    }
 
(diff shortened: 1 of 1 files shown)


0b5vr:
True! I'm sure this is an oversight.

It seems that VRM0 also sets `auto` as a default value. I will change the VRM0 side in a separated PR.
https://github.com/vrm-c/UniVRM/blob/07d98e2f1abc528d387f860d2224d0855b0d0b59/Assets/VRM/Runtime/FirstPerson/VRMFirstPerson.cs#L117-L119

0b5vr:
@mrxz The change has been released @ `v3.1.6`. Thank you again for the contribution!

https://github.com/pixiv/three-vrm/releases/tag/v3.1.6

PR #1525, opened 2024-11-07
fix: VRMAnimationLoaderPlugin, Improve hips position warning

0b5vr:
hips position is approximately zero -> y component of hips position is approximately zero or below

See the discussion also: https://github.com/pixiv/three-vrm/discussions/1524


ke456-png (approved):
Looks good.

PR #1531, opened 2024-11-15
ci: Bump some of GitHub actions

0b5vr:
download-artifact and upload-artifact will be deprecated on 12/5.

simply changing to v4 should do the job.


0b5vr:
artifacts working properly. merging

PR #1535, opened 2024-11-19
Add `VRMUtils.combineSkeletons` for reducing the number of `THREE.Skeleton`

mrxz:
As mentioned in #1532, here's a PR that introduces a utility combining `THREE.Skeleton` instances. The `combineSkeletons` function attempts to combine the skeletons of skinned meshes where possible. The fewer skeletons there are, the fewer computations have to be performed to compute and upload the bone matrices (see [Skeleton.js#L127-L153](https://github.com/mrdoob/three.js/blob/dev/src/objects/Skeleton.js#L127-L153)).

The following table shows the number of bones and the number of skeletons. For example, the `VRM1_Constraint_Twist_Sample.vrm` has 3 Skeletons, one with 83 bones and two with 137, so 357 bones (3 skeletons).
| Avatar | Number of bones (# skeletons) | After `removeUnnecessaryJoints` | After `combineSkeletons` |
|-----------|:-------------------------:|:-----------------------------------------------:|:------------------------------------:|
|VRM1_Constraint_Twist_Sample.vrm | 357 (3) | 174 (13) | 130 (2) |
|AvatarSample_C.vrm| 339 (3) | 132 (12) | 71 (1)|
|Zonko_VRM_221128_ps.vrm| 1647 (9) | 291 (21) | 133 (1) |
|AKAI Original VRM by Shugan.vrm| 3179 (17) | 376 (18) | 141 (1) |

As can be seen the `removeUnnecessaryJoints` can greatly reduce the number of bones, but does so by creating a unique `Skeleton` per skinned mesh. With `combineSkeletons` it instead tries to create as few skeletons as possible for all the skinned meshes.

Only one of the two utilities will be needed, so if accepted it's probably a good idea to remove (or deprecate) `removeUnnecessaryJoints`. There might be users who depend on the fact that each `SkinnedMesh` had its own `Skeleton` instance when using `removeUnnecessaryJoints`.

### DIFF
--- a/packages/three-vrm/src/VRMUtils/combineSkeletons.ts
+++ b/packages/three-vrm/src/VRMUtils/combineSkeletons.ts
@@ -1,0 +1,121 @@
+import * as THREE from 'three';
+
+/**
+ * Traverses the given object and combines the skeletons of skinned meshes.
+ *
+ * Each frame the bone matrices are computed for every skeleton. Combining skeletons
+ * reduces the number of calculations needed, improving performance.
+ *
+ * @param root Root object that will be traversed
+ */
+export function combineSkeletons(root: THREE.Object3D): void {
+  const skinnedMeshes = new Set<THREE.SkinnedMesh>();
+  const geometryToSkinnedMesh = new Map<THREE.BufferGeometry, THREE.SkinnedMesh>();
+
+  // Traverse entire tree and collect skinned meshes
+  root.traverse((obj) => {
+    if (obj.type !== 'SkinnedMesh') {
+      return;
+    }
+
+    const skinnedMesh = obj as THREE.SkinnedMesh;
+
+    // Check if the geometry has already been encountered
+    const previousSkinnedMesh = geometryToSkinnedMesh.get(skinnedMesh.geometry);
+    if (previousSkinnedMesh) {
+      // Skinned meshes that share their geometry with other skinned meshes can't be processed.
+      // The skinnedMeshes already contain previousSkinnedMesh, so remove it now.
+      skinnedMeshes.delete(previousSkinnedMesh);
+    } else {
+      geometryToSkinnedMesh.set(skinnedMesh.geometry, skinnedMesh);
+      skinnedMeshes.add(skinnedMesh);
+    }
+  });
+
+  // Prepare new skeletons for the skinned meshes
+  const newSkeletons: Array<{ bones: THREE.Bone[]; boneInverses: THREE.Matrix4[]; meshes: THREE.SkinnedMesh[] }> = [];
+  skinnedMeshes.forEach((skinnedMesh) => {
+    const skeleton = skinnedMesh.skeleton;
+
+    // Find suitable skeleton
+    let newSkeleton = newSkeletons.find((candidate) => skeletonMatches(skeleton, candidate));
+    if (!newSkeleton) {
+      newSkeleton = { bones: [], boneInverses: [], meshes: [] };
+      newSkeletons.push(newSkeleton);
+    }
+
+    // Add skinned mesh to the new skeleton
+    newSkeleton.meshes.push(skinnedMesh);
+
+    // Determine bone index mapping from skeleton -> newSkeleton
+    const boneIndexMap: number[] = skeleton.bones.map((bone) => newSkeleton.bones.indexOf(bone));
+
+    // Update skinIndex attribute
+    const geometry = skinnedMesh.geometry;
+    const attribute = geometry.getAttribute('skinIndex');
+    const weightAttribute = geometry.getAttribute('skinWeight');
+
+    for (let i = 0; i < attribute.count; i++) {
+      for (let j = 0; j < attribute.itemSize; j++) {
+        // check bone weight
+        const weight = weightAttribute.getComponent(i, j);
+        if (weight === 0) {
+          continue;
+        }
+
+        const index = attribute.getComponent(i, j);
+
+        // new skinIndex buffer
+        if (boneIndexMap[index] === -1) {
+          boneIndexMap[index] = newSkeleton.bones.length;
+          newSkeleton.bones.push(skeleton.bones[index]);
+          newSkeleton.boneInverses.push(skeleton.boneInverses[index]);
+        }
+
+        attribute.setComponent(i, j, boneIndexMap[index]);
+      }
+    }
+
+    attribute.needsUpdate = true;
+  });
+
+  // Bind new skeleton to the meshes
+  for (const { bones, boneInverses, meshes } of newSkeletons) {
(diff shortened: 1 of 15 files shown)


packages/three-vrm/src/VRMUtils/combineSkeletons.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    const skinnedMesh = obj as THREE.SkinnedMesh;
+
+    // Check if the geometry has already been encountered
+    const previousSkinnedMesh = geometryToSkinnedMesh.get(skinnedMesh.geometry);
+    if (previousSkinnedMesh) {
+      // Skinned meshes that share their geometry with other skinned meshes can't be processed.
+      // The skinnedMeshes already contain previousSkinnedMesh, so remove it now.
+      skinnedMeshes.delete(previousSkinnedMesh);
```
0b5vr, line 28: I doubt this code won't work correctly if we encounter an odd count (3, 5, ...) of skinned meshes that shares a same geometry? 50% confidence

0b5vr:
Could we actually replace the existing `removeUnnecessaryJoints` of our examples to `combineSkeletons`?

packages/three-vrm/src/VRMUtils/combineSkeletons.ts
0b5vr: Why we can't process such meshes by the way? Is it because we are going to modify the geometry skinIndex to match to the reorganized skeleton?

mrxz: Actually it should work reliably, as the mesh is removed from `skinnedMeshes`, but still in `geometryToSkinnedMesh`, which is used to check if the geometry has been seen before. So in case of 3 `SkinnedMeshes` with the same geometry, you'd get:

1. `previousSkinnedMesh = undefined` => `skinnedMeshes.add(skinnedMesh)`
2.  `previousSkinnedMesh = <first mesh>` => `skinnedMeshses.delete(previousMesh)`
3.  `previousSkinnedMesh = <first mesh>` => `skinnedMeshses.delete(previousMesh)` (no-op)

> Why we can't process such meshes by the way? Is it because we are going to modify the geometry skinIndex to match to the reorganized skeleton?

Correct, the `skinIndex` will be changed which would cause issues when processing the next `SkinnedMesh` with that geometry. The problem gets even worse when the geometry is the same, but the skeleton is different. Neither cases are likely and I have yet to see any model with such a setup, but I thought it would be better to have this check as a safeguard, rather than to assume geometries aren't shared.

mrxz:
> Could we actually replace the existing `removeUnnecessaryJoints` of our examples to `combineSkeletons`?

I've replaced `removeUnnecessaryJoints` with `combineSkeletons` in the examples.

Also made a small adjustment for WebGPU, as it didn't work correctly when creating the `Skeleton` and adding bones afterwards. So now the `bones` and `boneInverses` are collected and afterwards turned into `Skeleton` instances.

0b5vr:
The function failed to skin correctly on one of our test models (I can't share the model). I will try to find out what causes the issue.

packages/three-vrm/src/VRMUtils/combineSkeletons.ts
0b5vr: > Actually it should work reliably, as the mesh is removed from `skinnedMeshes`, but still in `geometryToSkinnedMesh`, which is used to check if the geometry has been seen before.

Ah, I simply missaw that the code deletes the item from `skinnedMeshes`, not `geometryToSkinnedMesh`. Thank you for clarifying.

packages/three-vrm/src/VRMUtils/combineSkeletons.ts
```diff
@@ -0,0 +1,121 @@
...
+
+// https://github.com/mrdoob/three.js/blob/r170/test/unit/src/math/Matrix4.tests.js#L12
+function matrixEquals(a: THREE.Matrix4, b: THREE.Matrix4, tolerance?: number) {
+  tolerance = tolerance || 0.0001;
+  if (a.elements.length != b.elements.length) {
+    return false;
+  }
+
+  for (let i = 0, il = a.elements.length; i < il; i++) {
+    const delta = a.elements[i] - b.elements[i];
```
0b5vr: I got what is wrong:

```suggestion
    const delta = Math.abs(a.elements[i] - b.elements[i]);
```

0b5vr:
I just measured the performance, the GPU performance drastically improves on WebGLRenderer!

However, on WebGPURenderer, it seems that the performance doesn't change much.
I'm not sure why though; having the `experimentalSameBoneCounts` option like `removeUnnecessaryJoints` might improve the runtime performance as well as the initial shader compilation time. Low confidence.

packages/three-vrm/src/VRMUtils/combineSkeletons.ts
mrxz: Good find. Quite surprised to see this bug as the code was taken straight from the three.js unit tests.
For reference, I've opened a PR fixing the issue in Three.js: https://github.com/mrdoob/three.js/pull/29946

mrxz:
> I just measured the performance, the GPU performance drastically improves on WebGLRenderer!

That's great! Not sure why WebGPURenderer doesn't benefit as much from this change as WebGLRenderer. I haven't looked into the skinning of `WebGPURenderer` much, but it seems that it might create a buffer for the `boneMatrices` _per `SkinnedMesh`_ regardless if the `Skeleton` is shared or not.

0b5vr:
> I haven't looked into the skinning of WebGPURenderer much, but it seems that it might create a buffer for the boneMatrices per SkinnedMesh regardless if the Skeleton is shared or not.

I agree, I suspect that the buffers are created per `SkinnedMesh`. Here is a line and `device.createBuffer` is invoked many times for the same materices buffer.

https://github.com/mrdoob/three.js/blob/4cc2eb5f5cde14e844266521e8b3b1daf6767ae9/src/renderers/webgpu/utils/WebGPUBindingUtils.js#L263

0b5vr:
I tried implementing `experimentalSameBoneCounts`. however, the count of compiled shaders is the same when `experimentalSameBoneCounts` is disabled and enabled, so we don't need the `experimentalSameBoneCounts` feature. Checked by seeing the count of `renderer._nodes.nodeBuilderCache`.

0b5vr (approved):
I believe the feature is sweet enough to have in three-vrm as is. I appreciate the contribution very much ❤️

0b5vr:
@mrxz Do you have anything left to do in this PR? If not, I will merge this, put a deprecation warning in `removeUnnecessaryJoints`, and ship it as a minor bump.

mrxz:
Nothing left to do from my side, this PR is good to go.

PR #1539, opened 2024-11-25
Optimize spring bones computations

mrxz:
Spring bones tend to be quite performance intensive on the CPU side. I looked into ways to optimize them, while retaining their original behaviour. This PR consists of three commits each focussing on a different area. 

See the following table which shows the average duration a call to `vrm.update` took. The changes are stacked, so the `Collider shape offset` column contains all three optimizations.
| **Avatar**                       | **Before PR** | **Pre-compute order** | **World space computations** | **Collider shape offset** |
|----------------------------------|:----------:|:--------------------:|:----------------------------:|:-------------------------:|
| VRM1_Constraint_Twist_Sample.vrm |  658.0 µs  |       179.1 µs       |           164.3 µs           |          145.7 µs         |
| AvatarSample_C.vrm               |   1.0 ms   |       615.3 µs       |           473.8 µs           |          415.2 µs         |
| Zonko_VRM_221128_ps.vrm          |   3.2 ms   |       318.0 µs       |           269.1 µs           |          200.1 µs         |
| AKAI Original VRM by Shugan.vrm  |  161.7 µs  |       105.8 µs       |            92.4 µs           |          95.1 µs          |

### Pre-computing spring joint update order (a21e474893b917d0d96bfe3e69de66224c357ea4)
The order in which spring bones need to be updated can be pre-computed making the actual `update` a lot more straightforward and performant. Whenever a spring bone is added or removed, the order is recomputed. The following things are precomputed:

- The order in which the spring bones should be updated when taking all dependencies into account (`_sortedJoints`)
- The ancestors from the [lowest common ancestor](https://en.wikipedia.org/wiki/Lowest_common_ancestor) of all the spring bone roots to the spring bone roots (`_ancestors`)

### Computing stiffness and gravity in world space (2e147298e7ff9fa66035890f0e87ab5da2c3489a)
The `VRMSpringBoneJoint.update` method converted all forces into center space. This is required for inertia, but for both stiffness and gravity this isn't strictly needed. It saves a couple of matrix multiplications by handling these in world space instead.

Additionally some of the vector math could be simplified (e.g. `.add(v3.multiplyScalar(scalar))` -> `.addScaledVector(v3, scalar)`) slightly reducing the amount of operations required.

### Aligning collider position with collider shape offset (7449a0ddff4fda636ee4a834c6afc956b8d6b476)
The `VRMSpringBoneCollider` is a `THREE.Object3D` with an identity transform. The collision shapes contain an `offset`, which is used to compute the world position of the collider shape. But this is done each time `calculateCollision` is called, which can be many times per-frame, despite the actual world position of the collider not changing in between.

By setting the position of `VRMSpringBoneCollider` to the offset, the world position can be retrieved from the `worldMatrix` instead, ensuring it's only calculated once per collider per update. To further reduce the number of calculations done, the output vector (`target`) is now only updated in case the distance is negative. 

This did require some changes to the unit tests, as they don't use `VRMSpringBoneCollider` but test against the `VRMSpringBoneColliderShape` directly.

### Notes
For benchmarking and testing I created a setup where I loaded the VRM, called `vrm.update(FIXED_DELTA)` and compared the world positions and rotations of all joints to reference values based on running the same on `dev`. So for the cases and VRM models tested above, the output after one update was identical (with EPSILON=0.0001).

### DIFF
--- a/packages/three-vrm-springbone/src/VRMSpringBoneCollider.ts
+++ b/packages/three-vrm-springbone/src/VRMSpringBoneCollider.ts
@@ -10,9 +10,43 @@ export class VRMSpringBoneCollider extends THREE.Object3D {
    */
   public readonly shape: VRMSpringBoneColliderShape;
 
+  /**
+   * World space matrix for the collider shape used in collision calculations.
+   */
+  public readonly colliderMatrix = new THREE.Matrix4();
+
   public constructor(shape: VRMSpringBoneColliderShape) {
     super();
 
     this.shape = shape;
+  }
+
+  public updateWorldMatrix(updateParents: boolean, updateChildren: boolean): void {
+    super.updateWorldMatrix(updateParents, updateChildren);
+
+    updateColliderMatrix(this.colliderMatrix, this.matrixWorld, this.shape.offset);
+  }
+}
+
+/**
+ * Computes the colliderMatrix based on an offset and a world matrix.
+ * Equivalent to the following code when matrixWorld is an affine matrix:
+ * ```js
+ * out.makeTranslation(offset).premultiply(matrixWorld)
+ * ```
+ *
+ * @param colliderMatrix The target matrix to store the result in.
+ * @param matrixWorld The world matrix fo the collider object.
+ * @param offset Optional offset to the collider shape.
+ */
+function updateColliderMatrix(colliderMatrix: THREE.Matrix4, matrixWorld: THREE.Matrix4, offset?: THREE.Vector3) {
+  const me = matrixWorld.elements;
+
+  colliderMatrix.copy(matrixWorld);
+
+  if (offset) {
+    colliderMatrix.elements[12] = me[0] * offset.x + me[4] * offset.y + me[8] * offset.z + me[12];
+    colliderMatrix.elements[13] = me[1] * offset.x + me[5] * offset.y + me[9] * offset.z + me[13];
+    colliderMatrix.elements[14] = me[2] * offset.x + me[6] * offset.y + me[10] * offset.z + me[14];
   }
 }
--- a/packages/three-vrm-springbone/src/VRMSpringBoneJoint.ts
+++ b/packages/three-vrm-springbone/src/VRMSpringBoneJoint.ts
@@ -84,6 +84,26 @@ export class VRMSpringBoneJoint {
   private _worldSpaceBoneLength = 0.0;
 
   /**
+   * Set of dependencies that need to be updated before this joint.
+   */
+  public get dependencies(): Set<THREE.Object3D> {
+    const set = new Set<THREE.Object3D>();
+
+    const parent = this.bone.parent;
+    if (parent) {
+      set.add(parent);
+    }
+
+    for (let cg = 0; cg < this.colliderGroups.length; cg++) {
+      for (let c = 0; c < this.colliderGroups[cg].colliders.length; c++) {
+        set.add(this.colliderGroups[cg].colliders[c]);
+      }
+    }
+
+    return set;
+  }
+
+  /**
    * This springbone will be calculated based on the space relative from this object.
    * If this is `null`, springbone will be calculated in world space.
    */
--- a/packages/three-vrm-springbone/src/VRMSpringBoneJoint.ts
+++ b/packages/three-vrm-springbone/src/VRMSpringBoneJoint.ts
@@ -1,5 +1,4 @@
 import * as THREE from 'three';
-import { mat4InvertCompat } from './utils/mat4InvertCompat';
 import { Matrix4InverseCache } from './utils/Matrix4InverseCache';
 import type { VRMSpringBoneColliderGroup } from './VRMSpringBoneColliderGroup';
 import type { VRMSpringBoneJointSettings } from './VRMSpringBoneJointSettings';
@@ -13,7 +12,6 @@ const IDENTITY_MATRIX4 = new THREE.Matrix4();
 // 計算中の一時保存用変数（一度インスタンスを作ったらあとは使い回す）
 const _v3A = new THREE.Vector3();
 const _v3B = new THREE.Vector3();
-const _v3C = new THREE.Vector3();
 
 /**
  * A temporary variable which is used in `update`
@@ -23,16 +21,9 @@ const _worldSpacePosition = new THREE.Vector3();
 /**
  * A temporary variable which is used in `update`
  */
-const _centerSpacePosition = new THREE.Vector3();
-
-/**
- * A temporary variable which is used in `update`
- */
 const _nextTail = new THREE.Vector3();
 
-const _quatA = new THREE.Quaternion();
 const _matA = new THREE.Matrix4();
-const _matB = new THREE.Matrix4();
 
 /**
  * A class represents a single joint of a spring bone.
@@ -204,7 +195,7 @@ export class VRMSpringBoneJoint {
     }
 
     // copy the child position to tails
-    const matrixWorldToCenter = this._getMatrixWorldToCenter(_matA);
+    const matrixWorldToCenter = this._getMatrixWorldToCenter();
     this.bone.localToWorld(this._currentTail.copy(this._initialLocalChildPosition)).applyMatrix4(matrixWorldToCenter);
     this._prevTail.copy(this._currentTail);
 
@@ -224,7 +215,7 @@ export class VRMSpringBoneJoint {
     this.bone.matrixWorld.multiplyMatrices(this._parentMatrixWorld, this.bone.matrix);
 
     // Apply updated position to tail states
-    const matrixWorldToCenter = this._getMatrixWorldToCenter(_matA);
+    const matrixWorldToCenter = this._getMatrixWorldToCenter();
     this.bone.localToWorld(this._currentTail.copy(this._initialLocalChildPosition)).applyMatrix4(matrixWorldToCenter);
     this._prevTail.copy(this._currentTail);
   }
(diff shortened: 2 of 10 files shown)


packages/three-vrm-springbone/src/VRMSpringBoneJoint.ts
```diff
@@ -83,6 +83,26 @@ export class VRMSpringBoneJoint {
    */
   private _worldSpaceBoneLength = 0.0;
 
+  /**
+   * List of dependencies that need to be updated before this joint.
+   */
+  public get dependencies(): THREE.Object3D[] {
```
0b5vr: [SHOULD] Existing VRMConstraints already have signatures `get dependencies(): Set<THREE.Object3D>`. I want to align the interface unless it needs to be an array instead.

0b5vr: Could you explain why you moved this to the public property though? I don't think it's a bad idea, but it is still called only from `VRMSpringBoneManager`.

packages/three-vrm-springbone/src/utils/lowestCommonAncestor.ts
```diff
@@ -0,0 +1,25 @@
+import type * as THREE from 'three';
+
+/**
+ * Finds the lowest common ancestors of the given objects, if it exists.
+ * @param objects The objects to find the lowest common ancestor for.
+ */
+export function lowestCommonAncestor(objects: THREE.Object3D[]): THREE.Object3D | null {
```
0b5vr: Hadn't heard of Lowest Common Ancestor but it seems it's a pretty common term in graph theory 👍

packages/three-vrm-springbone/src/utils/lowestCommonAncestor.ts
```diff
@@ -0,0 +1,25 @@
...
+  const sharedAncestors = new Map<THREE.Object3D, number>();
+  for (const object of objects) {
+    let current: THREE.Object3D | null = object;
+    do {
+      if (!sharedAncestors.has(current)) {
+        sharedAncestors.set(current, 0);
+      }
+
+      const newValue = sharedAncestors.get(current)! + 1;
```
0b5vr:
```suggestion
      const newValue = (sharedAncestors.get(current) ?? 0) + 1;
```

does this do the same job?

packages/three-vrm-springbone/src/VRMSpringBoneManager.ts
```diff
@@ -80,135 +89,123 @@ export class VRMSpringBoneManager {
     this.deleteJoint(joint);
   }
 
-  public setInitState(): void {
+  /**
+   * Sorts the joints ensuring they are updated in the correct order taking dependencies into account.
+   */
+  private _sortJoints() {
```
0b5vr, line 139: [MUST] Put private methods below the public methods. that will also make the diff look better.

packages/three-vrm-springbone/src/VRMSpringBoneManager.ts
```diff
@@ -80,135 +89,123 @@ export class VRMSpringBoneManager {
...
+    const springBoneOrder: Array<VRMSpringBoneJoint> = [];
     const springBonesTried = new Set<VRMSpringBoneJoint>();
     const springBonesDone = new Set<VRMSpringBoneJoint>();
-    const objectUpdated = new Set<THREE.Object3D>();
-
+    const ancestors: Array<THREE.Object3D> = [];
```
0b5vr: [SHOULD] I prefer a set instead of an array if there is no specific reason.

packages/three-vrm-springbone/src/VRMSpringBoneManager.ts
```diff
@@ -80,135 +89,123 @@ export class VRMSpringBoneManager {
...
-      // update the springbone
-      this._processSpringBone(springBone, springBonesTried, springBonesDone, objectUpdated, (springBone) =>
-        springBone.update(delta),
-      );
-
-      // update children world matrices
-      // it is required when the spring bone chain is sparse
-      traverseChildrenUntilConditionMet(springBone.bone, (object) => {
+    const lca = lowestCommonAncestor(ancestors);
+    this._ancestors = [];
```
0b5vr, line 150: I'm sure this `this._ancestors` has to be an array since the 0th element should calculate ancestors' world matrices.

[SHOULD] However, the reason why the variable `this._ancestors` is needed is kinda unintuitive. It would be very appreciated if you give them a proper doc comment to help future us.
I believe it goes like "A list of ancestors of the entire SpringBone joints. Their world matrices have to be calculated before we update the entire SpringBone system. The first element is ~~"

packages/three-vrm-springbone/src/VRMSpringBoneJoint.ts
```diff
@@ -204,7 +195,7 @@ export class VRMSpringBoneJoint {
     }
 
     // copy the child position to tails
-    const matrixWorldToCenter = this._getMatrixWorldToCenter(_matA);
+    const matrixWorldToCenter = this._getMatrixWorldToCenter();
```
0b5vr, line 198: true, we don't need to copy matrices at all! 😂

packages/three-vrm-springbone/src/VRMSpringBoneJoint.ts
```diff
@@ -241,64 +232,42 @@ export class VRMSpringBoneJoint {
...
     // update prevTail and currentTail
-    matrixWorldToCenter = this._getMatrixWorldToCenter(_matA);
-
     this._prevTail.copy(this._currentTail);
-    this._currentTail.copy(_v3A.copy(_nextTail).applyMatrix4(matrixWorldToCenter));
+    this._currentTail.copy(_nextTail).applyMatrix4(this._getMatrixWorldToCenter());
 
     // Apply rotation, convert vector3 thing into actual quaternion
     // Original UniVRM is doing center unit calculus at here but we're gonna do this on local unit
-    const worldSpaceInitialMatrixInv = mat4InvertCompat(
```
0b5vr, line 273: I don't know why we still have `mat4InvertCompat` here... Thank you for deleting this.

0b5vr:
I reviewed the first two commits.
I've left several comments on the first one. I might miss the point so feel free to correct them.
The second one looks perfect to me.

I will review the third one tomorrow.

packages/three-vrm-springbone/src/VRMSpringBoneJoint.ts
mrxz: Initially I moved it as I wanted to cache the list of dependencies, since it caused a lot of memory allocations in the hot-path (`VMRSpringBoneManager.update`). The `VRMSpringBoneJoint` itself seemed like a good place to keep the cached list as it could also invalidate it when needed. But after pre-computing the joint order, it was no longer in the hot-path, so I removed the caching again.

In the end the move wasn't needed. But I like the consistency with the `VRMConstraints`, so I'll update the signature to return a `Set<THREE.Object3D>` as well.

packages/three-vrm-springbone/src/VRMSpringBoneManager.ts
mrxz: Correct, the `_ancestors` list is ordered such that their world matrices can be updated correctly. The first element is the lowest common ancestor for which we also need to make sure its ancestors have their world matrix updated.

The graph below is an example of a Scene graph and shows which nodes would be in the `_ancestors` list in that case:
```mermaid
graph TD
    S[Scene] --> AP[Object3D]
    S[Scene] --> B[Object3D]
    AP --> A[Object3D<br>ancestors#91;0#93;]
    A --> C[Object3D<br>ancestors#91;1#93;]
    A --> D[Object3D<br>ancestors#91;2#93;]
    C --> S1[Springbone 1]
    D --> S2[Springbone 2]
    D --> S3[Springbone 3]
  classDef ancestor fill:#AFA,stroke:gray
  classDef lancestor fill:#4D4,stroke:black
  classDef o3d fill:#EEE,stroke:gray
  class C ancestor
  class D ancestor
  class A lancestor
  class S o3d
  class AP o3d
  class B o3d
```

I've added a doc comment that should hopefully clarify things for future readers.

0b5vr: The doc comment looks perfect, thank you!

packages/three-vrm-springbone/src/VRMSpringBoneCollider.ts
```diff
@@ -14,5 +14,8 @@ export class VRMSpringBoneCollider extends THREE.Object3D {
     super();
 
     this.shape = shape;
+    if ('offset' in shape && (shape.offset as THREE.Vector3).isVector3) {
```
0b5vr: [SHOULD] how about defining `public offset?: THREE.Vector3;` in `VRMSpringBoneColliderShape` ?

0b5vr:
I'm kinda negative about the concept of the third commit. This change would prevent users from setting `offset` and `tail` after initializing, which is a breaking change in the first place.
We have yet to provide many editing features so far in three-vrm, but I still want to be conservative about making this change.

The save on the `length()` calculation looks great. Does this alone provide a performance improvement?

mrxz:
> I'm kinda negative about the concept of the third commit. This change would prevent users from setting offset and tail after initializing, which is a breaking change in the first place.

I agree, the commit indeed makes changing the colliders after initialization difficult. This isn't a problem when just loading an avatar and using it, but users might want to implement editing features.

I've changed it so that the `colliderMatrix` is computed in the `updateWorldMatrix` method of `VRMSpringBoneCollider` instead. This way the `offset` and `tail` can still be changed and the collider matrix is still only computed once per update. The `VRMSpringBoneManager` guarantees that `updateWorldMatrix` is called for the colliders before updating the spring bones. Let me know what you think.

packages/three-vrm-springbone/src/tests/VRMSpringBoneColliderShapeCapsule.test.ts
```diff
@@ -87,19 +87,19 @@ describe('VRMSpringBoneColliderShapeCapsule', () => {
       });
 
       it('must calculate a collision properly, object is near the tail', () => {
-        const colliderMatrix = new THREE.Matrix4();
+        const colliderMatrix = new THREE.Matrix4().makeTranslation(shape.offset);
         const objectPosition = new THREE.Vector3(3.0, 0.0, 0.0);
-        const objectRadius = 1.0;
+        const objectRadius = 2.0;
```
0b5vr, line 92: what is the intention of the change of `objectRadius` ?

packages/three-vrm-springbone/src/VRMSpringBoneCollider.ts
```diff
@@ -10,9 +10,44 @@ export class VRMSpringBoneCollider extends THREE.Object3D {
...
+  /**
+   * World space matrix for the collider shape used in collision calculations.
+   */
+  public readonly colliderMatrix = new THREE.Matrix4();
+
   public constructor(shape: VRMSpringBoneColliderShape) {
     super();
 
     this.shape = shape;
+    this.updateWorldMatrix(false, false);
```
0b5vr: why is the initial calculation needed?

packages/three-vrm-springbone/src/tests/VRMSpringBoneColliderShapeCapsule.test.ts
mrxz: The test was the only one that didn't report a collision (`dist > 0`). The output direction (`dir`) is now only normalized when there is a collision. So either the assertion on `dir` could be removed or the test changed to have a collision.

Changing the `objectRadius` seemed the easiest way to cause a collision to take place. Especially since the test is intended to test the tail of a capsule, it feels incomplete if it only checked the `dist`.

packages/three-vrm-springbone/src/VRMSpringBoneCollider.ts
mrxz: Is not needed, removed it. The `VRMSpringBoneManager` already guarantees that `updateWorldMatrix` is called. Thought it would be nice to ensure the `colliderMatrix` was correct right after initialization, but in practice it's value is outdated as soon as the collider is added to the scenegraph and positioned.

packages/three-vrm-springbone/src/tests/VRMSpringBoneColliderShapeCapsule.test.ts
0b5vr: I see. Thank you for the explanation!

0b5vr (approved):
Seems it's working correctly, and it's much more performant, especially on models with a lot of spring bones. Thank you for your contribution!

packages/three-vrm-springbone/src/VRMSpringBoneManager.ts
```diff
@@ -81,134 +98,122 @@ export class VRMSpringBoneManager {
...
 
     if (springBonesTried.has(springBone)) {
       if (!this._hasWarnedCircularDependency) {
-        console.warn('VRMSpringBoneManager: Circular dependency detected while updating springbones');
+        console.warn('VRMSpringBoneManager: Circular dependency detected');
         this._hasWarnedCircularDependency = true;
       }
       return;
     }
-    springBonesTried.add(springBone);
```
0b5vr, line 164: @mrxz We forgot to put this line to `_insertJointSort`!

I want to ask you why you deleted this line, just in case. Have you encountered any issues with having this line? If you simply forgot, it's all fine.

mrxz: @0b5vr I must have accidentality removed that line while moving code around, sorry. It should be added back in.

0b5vr: I see I see, thank you for the explanation!

0b5vr: A PR adds the line: https://github.com/pixiv/three-vrm/pull/1556#issue-2735402313

PR #1540, opened 2024-11-26
feat: put a deprecation warning to VRMUtils.removeUnnecessaryJoints

0b5vr:
This PR deprecates `VRMUtils.removeUnnecessaryJoints`.

In #1535, we got a superior function, `VRMUtils.combineSkeletons`, so we no longer have a purpose of using `removeUnnecessaryJoints`.

When using the function, it puts a warning that recommends the user to switch to `combineSkeletons`.
However, the implementation of `removeUnnecessaryJoints` is preserved in case the implementation of `combineSkeletons` causes a problem.
The function body will be removed in the next major version.


ke456-png (approved):
Looks good.

PR #1546, opened 2024-12-02
VRMUtils.combineMorphs

### LINKED ISSUE #1369: Missing head in a specific VRM model in specific Android devices
I encounter a strange issue with a VRM model. The VRM model that loads fine in desktop browsers. However, in some specific mobile devices (Android), the head of the VRM model went missing. I faced the issue in my ReactJS application. This issue does not happen on other VRM models.

Attached with a sample VRM model that has this issue: https://we.tl/t-XxxsvIAK1c

I tried reducing texture size, deleting everything except the head, but the head still doesn't load.

image

Any ideas what's the problem here?

0b5vr:
This will resolve https://github.com/pixiv/three-vrm/issues/1369

This PR will add `VRMUtils.combineMorphs`.

This function will prevent shader errors caused by the limitation of the number of morph targets, especially on mobile devices.
This might potentially improve the performance when using models with a lot of unnecessary morphs.

This PR also exposes `VRMExpression.binds` and adds `VRMExpression.deleteBind()`.
These were required to implement `VRMUtils.combineMorphs`.

### DIFF
--- a/packages/three-vrm/src/VRMUtils/combineMorphs.ts
+++ b/packages/three-vrm/src/VRMUtils/combineMorphs.ts
@@ -51,6 +51,11 @@ function combineMorph(
 }
 
 /**
+ * A map from expression names to a set of morph target binds.
+ */
+type NameBindSetMap = Map<string, Set<VRMExpressionMorphTargetBind>>;
+
+/**
  * Combine morph targets by VRM expressions.
  *
  * This function prevents crashes caused by the limitation of the number of morph targets, especially on mobile devices.
@@ -61,7 +66,7 @@ export function combineMorphs(vrm: VRMCore): void {
   const meshes = collectMeshes(vrm.scene);
 
   // Iterate over all expressions and check which morph targets are used
-  const meshBindListMap = new Map<THREE.Mesh, [string, VRMExpressionMorphTargetBind][]>();
+  const meshNameBindSetMapMap = new Map<THREE.Mesh, NameBindSetMap>();
 
   const expressionMap = vrm.expressionManager?.expressionMap;
   if (expressionMap != null) {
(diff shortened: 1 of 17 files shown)


0b5vr:
Since morph targets will be written into a data texture array, it don't have to be an interleaved buffer.

https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLMorphtargets.js#L62

packages/three-vrm/src/VRMUtils/combineMorphs.ts
```diff
@@ -0,0 +1,153 @@
...
+    weightSum = 1.0;
+  } else {
+    for (const bind of binds) {
+      weightSum += bind.weight;
+    }
+  }
+
+  for (const bind of binds) {
+    const src = positionAttributes[bind.index];
+    const weight = bind.weight / weightSum;
```
yue4u, line 40: MEMO:  

1. `bind.weight` is checked before adding to the bindset
2. `binds.length` always > 0

So it's not possible for `weightSum` to be kept 0 in order to cause zero division.

packages/three-vrm/src/VRMUtils/combineMorphs.ts
```diff
@@ -0,0 +1,153 @@
...
+
+      const nameBindSetMap = new Map<string, Set<VRMExpressionMorphTargetBind>>();
+      for (const [name, bind] of bindList) {
+        let set = nameBindSetMap.get(name);
+        if (set == null) {
+          set = new Set<VRMExpressionMorphTargetBind>();
+          nameBindSetMap.set(name, set);
+        }
+        set.add(bind);
+      }
```
yue4u: Question: `nameBindSetMap` is built from `bindList`, so why not make `meshBindListMap` typeof `Map<THREE.Mesh, Map<string, Set<VRMExpressionMorphTargetBind>>>`?

0b5vr: true. I'm gonna do this.

0b5vr: true. I'm gonna do this.

yue4u (approved):

PR #1553, opened 2024-12-10
Improve performance of combineMorphs

0b5vr:
This PR tries to improve the initial stutter when combineMorphs is used after the loading.

- [prevent cloning morph attributes.](https://github.com/pixiv/three-vrm/commit/850c04faf809feaecafa1d76bef9ed05ca609980)
- [if there is only one morph target and the weight is 1.0, we can use the original as-is.](https://github.com/pixiv/three-vrm/commit/2dfda5ce9db96bdf2f810eb93c58aaa820f71b8d)

### DIFF
--- a/packages/three-vrm/src/VRMUtils/combineMorphs.ts
+++ b/packages/three-vrm/src/VRMUtils/combineMorphs.ts
@@ -21,9 +21,17 @@ function collectMeshes(scene: THREE.Group): Set<THREE.Mesh> {
 
 function combineMorph(
   positionAttributes: (THREE.BufferAttribute | THREE.InterleavedBufferAttribute)[],
-  binds: Iterable<VRMExpressionMorphTargetBind>,
+  binds: Set<VRMExpressionMorphTargetBind>,
   morphTargetsRelative: boolean,
-): THREE.BufferAttribute {
+): THREE.BufferAttribute | THREE.InterleavedBufferAttribute {
+  // if there is only one morph target and the weight is 1.0, we can use the original as-is
+  if (binds.size === 1) {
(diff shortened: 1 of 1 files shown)


packages/three-vrm/src/VRMUtils/combineMorphs.ts
```diff
@@ -108,14 +116,18 @@ export function combineMorphs(vrm: VRMCore): void {
       continue;
     }
 
+    // prevent cloning morph attributes
+    const originalMorphAttributes = mesh.geometry.morphAttributes;
+    mesh.geometry.morphAttributes = {};
+
     const geometry = mesh.geometry.clone();
```
0b5vr, line 123: Here, cloning takes many CPU resources if the model has many morph attributes. the L121 prevents cloning morph attributes.

packages/three-vrm/src/VRMUtils/combineMorphs.ts
```diff
@@ -21,9 +21,17 @@ function collectMeshes(scene: THREE.Group): Set<THREE.Mesh> {
...
   morphTargetsRelative: boolean,
-): THREE.BufferAttribute {
+): THREE.BufferAttribute | THREE.InterleavedBufferAttribute {
+  // if there is only one morph target and the weight is 1.0, we can use the original as-is
+  if (binds.size === 1) {
+    const bind = binds.values().next().value!;
+    if (bind.weight === 1.0) {
+      return positionAttributes[bind.index];
+    }
+  }
```
0b5vr, line 33: Pretty straightforward improvement

ke456-png (approved):
Thanks!

PR #1554, opened 2024-12-10
feat: `VRMUtils.precompileShaders`

0b5vr:
This PR introduces `VRMUtils.precompileShaders`.

`precompileShaders` is a function to time slice the shader compilations to prevent the main thread from being blocked.

This function only supports WebGLRenderer at the moment.

### Points need review

- [ ] Is it nice to have in three-vrm?
    - We have a lot of stuff inside `VRMUtils` lately,,,
- [ ] Is the signature name appropriate?
    - Related APIs: `WebGLRenderer.compile()`, `WebGLRenderer.compileAsync()`, `WebGLRenderer.initTexture()`
    - Candidates: `precompileShaders`, `compileShaders`, `compileShadersAsync`, `precompile`, `compile`, `timeslicedCompile`, `timeslicedCompileAsync`, `prepareMaterials`, `prepareMaterialsAsync`, ...
        - It seems this is the first time we expose functions or methods that return `Promise` other than loader plugins.

### DIFF
--- a/packages/three-vrm/examples/dnd.html
+++ b/packages/three-vrm/examples/dnd.html
@@ -76,7 +76,7 @@
 
 					url,
 
-					( gltf ) => {
+					async ( gltf ) => {
 
 						const vrm = gltf.userData.vrm;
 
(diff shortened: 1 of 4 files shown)


0b5vr:
We are reluctant to merge this since this is a feature that each application requires their specific interfaces / implementations by demand

PR #1556, opened 2024-12-12
fix: SpringBone, forgot to add springBones to springBonesTried when sorting

0b5vr:
This fixes the `Maximum call stack size exceeded` exception happens when we load a model with a circular dependency in the spring bone system.

See: https://github.com/pixiv/three-vrm/pull/1539#discussion_r1881647521


ke456-png (approved):
Nice!

PR #1564, opened 2024-12-25
feat: Make it able to load VRMA without specVersion

0b5vr:
VRMAs exported from UniVRM prior to v0.120 doesn't have specVersion

See: https://github.com/vrm-c/UniVRM/pull/2251

Original discussion: https://github.com/pixiv/three-vrm/discussions/1563


ke456-png (approved):
Thanks!

PR #1569, opened 2025-01-14
feat: Fix combineSkeletons

0b5vr:
This PR fixes `combineSkeletons`.

There is a model that uses the same skinIndex accessor but not the same skinWeight accessors between meshes.
This PR updates the implementation of attributeUsedIndexSetMap to use both skinIndex and skinWeight accesstors.

The same skinIndex accessor could also be used between different bone sets.
Clone the bone set if the skinIndex accessor is the same but the bone set is different.

### DIFF
--- a/packages/three-vrm/src/VRMUtils/combineSkeletons.ts
+++ b/packages/three-vrm/src/VRMUtils/combineSkeletons.ts
@@ -265,7 +265,7 @@ function collectSkinIndexAttrs(
   >();
 
   for (const mesh of meshes) {
-    let skinIndexAttr = mesh.geometry.getAttribute('skinIndex');
+    const skinIndexAttr = mesh.geometry.getAttribute('skinIndex');
 
     // Get or create a map for the skin index attribute
     let newSkinIndexBonesMap = skinIndexNewSkinIndexBonesMapMap.get(skinIndexAttr);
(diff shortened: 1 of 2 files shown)


packages/three-vrm/src/VRMUtils/combineSkeletons.ts
```diff
@@ -222,3 +235,62 @@ function matrixEquals(a: THREE.Matrix4, b: THREE.Matrix4, tolerance?: number) {
...
+ */
+function collectSkinIndexAttrs(
+  meshes: Iterable<THREE.SkinnedMesh>,
+): Set<[THREE.BufferAttribute | THREE.InterleavedBufferAttribute, THREE.Bone[]]> {
+  const skinIndexBonesPairSet = new Set<[THREE.BufferAttribute | THREE.InterleavedBufferAttribute, THREE.Bone[]]>();
+
+  // Collect skin index attributes
+  // skinIndex attribute might be shared among different bone sets
+  // If there are multiple bone sets that share the same skinIndex attribute, clone the attribute
+  const skinIndexNewSkinIndexBonesMapMap = new Map<
```
0b5vr, line 262: If there is a skin index accessor that is used with two or more different bone sets, you can check by seeing inside the `skinIndexNewSkinIndexBonesMapMap`

packages/three-vrm/src/VRMUtils/combineSkeletons.ts
```diff
@@ -222,3 +235,62 @@ function matrixEquals(a: THREE.Matrix4, b: THREE.Matrix4, tolerance?: number) {
...
+      skinIndexBonesPairSet.add([skinIndexAttr, mesh.skeleton.bones]);
+      continue;
+    }
+
+    // Check if the bone set is already registered
+    for (const bones of newSkinIndexBonesMap.values()) {
+      if (arrayEquals(bones, mesh.skeleton.bones)) {
+        continue;
+      }
+    }
```
ke456-png: There is a problem with the scope of the `continue`.

ke456-png (changes requested):

packages/three-vrm/src/VRMUtils/combineSkeletons.ts
0b5vr: oh no

0b5vr: fixed @ https://github.com/pixiv/three-vrm/pull/1569/commits/902e6d83a4048a1e7a6bf4dfdb51252d7f8eb456

I also forgot to set the new attribute when we found an already registered skin index attribute 😅

ke456-png (approved):
Thanks!

PR #1588, opened 2025-03-05
New docs with monorepo support

0b5vr:
This PR introduces the new docs page with the monorepo support.

image

Existing docs for each package still exist and can be generated by running the `docs-legacy` command. These docs redirect to the new URL.
It consumes a long time to generate. We might want to remove it in the future.

Also, existing `/docs/` in the root level is moved to `/guides/`.

See details for the each commit.


typedoc-redirect-legacy-docs-plugin.mjs
```diff
@@ -0,0 +1,78 @@
...
+ * @returns The new path
+ */
+function replacePath(path) {
+  // Item pages
+  // Before: /packages/three-vrm-animation/docs/classes/VRMAnimation.html
+  // After: /docs/classes/three-vrm-materials-mtoon.MToonMaterial.html
+  const matchItems = path.match(/packages\/(.+)\/docs\/(classes|interfaces|functions|types|variables)\/(.+)/);
+  if (matchItems != null) {
+    const [_, packageName, kind, item] = matchItems;
+    return `../../../docs/${kind}/${packageName}.${item}`;
```
yue4u: The relative path `../../../` does not work for nested docs like `/packages/three-vrm/docs/variables/VRMExpressionMaterialColorType-1.html`. Maybe using `/docs/` is better?

yue4u (commented):
IMO the new doc is much better and keeping redirects is good!

typedoc-redirect-legacy-docs-plugin.mjs
0b5vr: Oops that's true. will change it to simply use `/docs/...`.

0b5vr: no that wouldn't work because the document root will be `https://pixiv.github.io/three-vrm/docs/` on GitHub Pages. I would try setting the document root using environment variable.

0b5vr: make it use an environment variable `DOCS_ROOT`.

https://github.com/pixiv/three-vrm/pull/1588/commits/8f9f9ae309409bc2ce1badeef1df90cf9b0dc606

yue4u (approved):
LGTM

PR #1615, opened 2025-04-02
fix: add undefined check.

MtBlue81:
I get the following error when reading a VRM file.

CleanShot 2025-04-02 at 09 34 44

The caller is only from the GLTFMTonMaterialParamsAssignHelper and is passed undefined.
The type of this._materialParams is [MToonMaterialParameters](https://github.com/pixiv/three-vrm/blob/9b3bd2c536d168fe2873e7b613817a9ebce0ab31/packages/three-vrm-materials-mtoon/src/MToonMaterialParameters.ts#L5), but Texture-related properties are optional.

In this PR, I added a check for undefined assuming it is optional. (use "! = null" to match the surrounding implementations)

### DIFF
--- a/packages/three-vrm-materials-mtoon/src/GLTFMToonMaterialParamsAssignHelper.ts
+++ b/packages/three-vrm-materials-mtoon/src/GLTFMToonMaterialParamsAssignHelper.ts
@@ -52,7 +52,7 @@ export class GLTFMToonMaterialParamsAssignHelper {
       if (texture != null) {
         await this._parser.assignTexture(this._materialParams, key, texture);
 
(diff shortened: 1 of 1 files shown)


0b5vr:
The `this._materialParams[key]` should be assigned at the `assignTexture`. I suspect that the glTF has an error. However, `assignTexture` doesn't give an error even if the texture cannot be found. I pretty much agree with the change, but I want to see the reproducer first.

https://github.com/mrdoob/three.js/blob/f230fe9147bfbf9c689a8d62b72d757395a513c9/examples/jsm/loaders/GLTFLoader.js#L3545

PR #1616, opened 2025-04-03
feat: fix the combineSkeletons issue when a same geometry is shared between meshes

0b5vr:
This PR fixes `combineSkeleton` when a same geometry is shared between meshes.

I simply cloned the geometry when the shared geometry is found.  
Since performing deep clone using `BufferGeometry#clone` affects the runtime performance, I had to implement `shallowCloneBufferGeometry`.

### DIFF
--- a/packages/three-vrm/src/VRMUtils/combineSkeletons.ts
+++ b/packages/three-vrm/src/VRMUtils/combineSkeletons.ts
@@ -11,15 +11,26 @@ import * as THREE from 'three';
 export function combineSkeletons(root: THREE.Object3D): void {
   const skinnedMeshes = collectSkinnedMeshes(root);
 
+  /** A set of geometries in the given {@link root}. */
+  const geometries = new Set<THREE.BufferGeometry>();
+  for (const mesh of skinnedMeshes) {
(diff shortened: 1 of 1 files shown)


packages/three-vrm/src/VRMUtils/combineSkeletons.ts
```diff
@@ -287,3 +300,42 @@ class ObjectIndexDispatcher<T> {
...
+  clone.name = geometry.name;
+
+  clone.setIndex(geometry.index);
+
+  for (const [name, attribute] of Object.entries(geometry.attributes)) {
+    clone.setAttribute(name, attribute);
+  }
+
+  for (const [name, morphAttribute] of Object.entries(geometry.morphAttributes)) {
+    clone.morphAttributes[name] = morphAttribute.concat();
```
yue4u, line 323: Why do we need concat here? To clone the attribute?

0b5vr: I simply wanted to create a cloned list of attributes while the reference of the attribute is preserved to be same.

yue4u (approved):

PR #1649, opened 2025-05-08
fix: Correct humanoid FBX root motion retargeting

### LINKED ISSUE #1643: VRM model sinks into floor after animation switch due to Mixamo hips position offsets
I'm using the provided loadMixamoAnimation.js utility to retarget Mixamo FBX animations onto a VRM model in Three.js. Animations play correctly, but when switching between them like this:
```js
async function loadAnimation() {
    if (!selectedAnimation.value || !currentVrm) return;

    const clip = await loadMixamoAnimation(
        `/animations/${selectedAnimation.value}`,
        currentVrm
    );

    const newAction = currentMixer.clipAction(clip);
    newAction.reset().play();

    if (currentAction && currentAction !== newAction) {
        currentAction.crossFadeTo(newAction, 0.3, true);
    }

    currentAction = newAction;
}
```

…the character gradually sinks lower into the ground with every animation switch, eventually clipping significantly into the floor.

After debugging, I found that this issue is caused by the hips.position track in the Mixamo animations. If I remove the hips position track entirely during retargeting, the model stays at a consistent height but the animation looks broken due to missing hip motion.

I've tried various normalization strategies but couldn’t find a solution that both:

✅ Keeps natural hip movement
❌ Prevents downward drifting over time

It seems like the absolute Y-offset in the hips.position track gets reapplied on every animation switch, and VRM’s root position doesn't reset, causing cumulative offset.

ke456-png:
Fix #1643.

Root motion retargeting was affected by current Hips world position. Modified to reference Humanoid.NormalizedRestPose, similar to VRMAnimation.

### DIFF
--- a/packages/three-vrm-core/examples/humanoidAnimation/loadMixamoAnimation.js
+++ b/packages/three-vrm-core/examples/humanoidAnimation/loadMixamoAnimation.js
@@ -25,9 +25,7 @@ export function loadMixamoAnimation( url, vrm ) {
 
 		// Adjust with reference to hips height.
 		const motionHipsHeight = asset.getObjectByName( 'mixamorigHips' ).position.y;
-		const vrmHipsY = vrm.humanoid?.getNormalizedBoneNode( 'hips' ).getWorldPosition( _vec3 ).y;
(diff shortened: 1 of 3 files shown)


0b5vr (approved):
Thank you!

PR #1654, opened 2025-05-13
fix: three-vrm-animation createVRMAnimationClip, fix fallback logic when VRMLookAtQuaternionProxy have no name

### LINKED ISSUE #1516: Fallback logic for `VRMLookAtQuaternionProxy.name` doesn't execute
I got tripped up when I tried adding a `VRMLookAtQuaternionProxy` to my VRM scene without manually setting its `.name`. I didn't see a warning message indicating what went wrong.

https://github.com/pixiv/three-vrm/blob/6e5a70a45b85041ad1d87552d2975b24129e68ff/packages/three-vrm-animation/src/createVRMAnimationClip.ts#L133-L140

The above logic attempts to set a default name if one is not manually set and warn the user to set it themselves.

However the logic here checks for a `null` name, when an unset name actually contains an empty string `''`. This is reflected in the type of `proxy.name` which is `string`. This means the fallback logic and helpful warning message don't execute.

I suggest changing this to check for empty string instead of null.

0b5vr:
This resolves #1516

`Object3D#name` can't be null or undefined since the constructor already assign `''`


ke456-png (approved):
Thanks!

PR #1739, opened 2025-10-02
Enable OIDC Trusted Publishing

0b5vr:
- bump lerna to v9.0.0
  - lerna supports OIDC Trusted Publishing since v9.0.0
  - See: https://github.com/lerna/lerna/releases/tag/v9.0.0
- remove NPM_TOKEN in favor of OIDC Trusted Publishing
  - permissions (contents, id-token) have already been set up correctly, so I just removed the NPM_TOKEN section
  - See: https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/

I have to setup the Trusted Publisher settings on the npm side next.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "globals": "^16.0.0",
     "husky": "^9.1.5",
     "jest": "^29.7.0",
-    "lerna": "^8.1.8",
+    "lerna": "^9.0.0",
     "lint-staged": "^16.0.0",
     "patch-package": "^8.0.0",
     "prettier": "^3.3.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "globals": "^16.0.0",
     "husky": "^9.1.5",
     "jest": "^29.7.0",
-    "lerna": "^8.1.8",
+    "lerna": "^9.0.0",
```
yue4u, line 37: breaking for node 18 EOL https://github.com/lerna/lerna/releases/tag/v9.0.0

0b5vr: yep, we're already using node 22 so it should be fine

yue4u (commented):
LGTM

0b5vr:
Just tried publishing `v3.4.3-alpha.0` and is working right. First try!

PR #1770, opened 2025-12-25
test, deps: migrate from jest to vitest

0b5vr:
much faster tests!

ref, custom matcher: https://vitest.dev/blog/vitest-3-2.html#the-new-multi-purpose-matchers-type


vitest.config.mjs
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'vitest/config';
```
yue4u, line 1: Maybe we can use vitest workspace(projects) support to avoid duplicate vitest config?

https://vitest.dev/guide/projects

0b5vr: tried @ https://github.com/pixiv/three-vrm/pull/1770/commits/a62c7a9e6d4042bc57634c963c84a5c7b12d2393

ke456-png (approved):
Looks good.

PR #1840, opened 2026-05-13
fix: VRMSpringBoneLoaderPlugin, VRM0 import, improve error handling

0b5vr:
VRMSpringBoneLoaderPlugin, VRM0 import, improve error handling

VRM0 exported from old UniVRM versions might contain `-1` to index nodes from spring bone groups

This no longer bails when the model has invalid spring bone settings. Tried to make the blast radius as small as possible.
See each commit for details.

### DIFF
--- a/packages/three-vrm-springbone/src/VRMSpringBoneLoaderPlugin.ts
+++ b/packages/three-vrm-springbone/src/VRMSpringBoneLoaderPlugin.ts
@@ -291,6 +291,12 @@ export class VRMSpringBoneLoaderPlugin implements GLTFLoaderPlugin {
     const colliderGroups = schemaSecondaryAnimation.colliderGroups?.map(
       (schemaColliderGroup): VRMSpringBoneColliderGroup => {
         const node = threeNodes[schemaColliderGroup.node!];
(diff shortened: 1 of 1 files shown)


packages/three-vrm-springbone/src/VRMSpringBoneLoaderPlugin.ts
```diff
@@ -180,23 +180,25 @@ export class VRMSpringBoneLoaderPlugin implements GLTFLoaderPlugin {
...
-            console.warn(
-              `VRMSpringBoneLoaderPlugin: The colliderGroup #${iColliderGroup} attempted to use a collider #${iCollider} but not found`,
-            );
-            return [];
-          }
+            if (col == null) {
+              console.warn(
+                `VRMSpringBoneLoaderPlugin: The collider group #${iColliderGroup} attempted to reference a collider #${iCollider} but not found. Skipping the collider`,
+              );
+              return [];
```
ke456-png: I think this should return null.
```suggestion
              return null;
```

ke456-png (approved):
Thanks!

PR #1841, opened 2026-05-13
fix: MToon loader, improve texture assignment error handling

0b5vr:
MToon loader, improve texture assignment error handling

It bailed when a loaded model contains corrupted textures, since assigned texture becomes null. This commit will make it output a warning instead of bailing


ke456-png (approved):
Looks good.

PR #1842, opened 2026-06-04
fix: three-vrm-core, humanoid, Handle negative node indices (-1) in VRM0.0 humanoid bones

0b5vr:
VRM0.0 can contain -1 as a node index, which is invalid. Found at least in UniVRM-0.61.1.


ke456-png:
I think error handling is needed here too.
https://github.com/pixiv/three-vrm/blob/159bc350e5df5f07ef0b4e09dd28abcf74ba3942/packages/three-vrm-core/src/humanoid/VRMHumanoidLoaderPlugin.ts#L178

ke456-png (approved):

PR #1852, opened 2026-07-08
fix: VRMSpringBoneLoaderPlugin, skip springs without joints

0b5vr:
Some VRMC_springBone assets can contain a spring entry without `joints`. Such a spring is invalid but we should go permissive. Show a warning and skip that spring instead of bailing out the entire loader.

Related: 806d70d4


ke456-png (approved):
Thanks!

